Issue #51, opened 2014-05-11
Utilize pagination headers


NARKOZ:

NARKOZ:
Implemented in #134

PR #85, opened 2014-11-03
Utilize pagination headers

walquis:
Takes a swing at issue #51.  Before enhancing, I initially added monkey-patches to accommodate my migration of repos from github to gitlab, then decided to make it a Real Change, then noticed #51 was already created; I added some tests for the "/users endpoint", made them pass, and replaced the rest of the endpoints that needed "paginate".  I built a local version of the gem for my testing, threw away my monkey-patches and integrated the paginating version of the Gitlab gem back into my migration scripts.  (The "/projects" endpoint also got a workout at this point).  The existing tests still pass.

NOTE: These changes have NOT been tested via direct invocation of the Gitlab gem's command-line interface, but only via unit tests and calls via inclusion into my scripts).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,3 +18,4 @@ tmp
 .idea
 .ruby-version
 .ruby-gemset
+vendor
--- a/lib/gitlab.rb
+++ b/lib/gitlab.rb
@@ -31,7 +31,7 @@ module Gitlab
   #
   # @return [Array<Symbol>]
   def self.actions
-    hidden = /endpoint|private_token|user_agent|sudo|get|post|put|\Adelete\z|validate|set_request_defaults/
+    hidden = /endpoint|private_token|user_agent|sudo|auto_paginate|get|post|put|\Adelete\z|validate|set_request_defaults/
     (Gitlab::Client.instance_methods - Object.methods).reject {|e| e[hidden]}
   end
 end
--- a/lib/gitlab/client.rb
+++ b/lib/gitlab/client.rb
@@ -14,5 +14,6 @@ module Gitlab
     include Snippets
     include SystemHooks
     include Users
+    include Helpers
   end
 end
--- a/lib/gitlab/client/branches.rb
+++ b/lib/gitlab/client/branches.rb
@@ -12,7 +12,7 @@ class Gitlab::Client
     # @option options [Integer] :per_page The number of results per page.
     # @return [Array<Gitlab::ObjectifiedHash>]
     def branches(project, options={})
-      get("/projects/#{project}/repository/branches", :query => options)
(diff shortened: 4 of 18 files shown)


StephenOTT:
Have we looked at the way the GitHub Ruby API Auto Pagination works?

NARKOZ:
@StephenOTT Thanks, will look.

e-sabelhaus:
Is there anything which needs to be done to merge this functionality into the gem? I would love to see this made available, and will help out if there is anything to do.

StephenOTT:
@e-sabelhaus if you need this right away, a simple monkey wrench way is to just build a loop that calls the API 1 page at a time (with the per page set to 100), when the returned array is nil, then you know you have reached the end of the data. 

Example of you have 5 issues and you call the issues API with per page set to 1. Then you start with page => 1 and increase by 1 per loop iteration.  When you hit page 4, the API will return nil and you will know you hit the end of the data. 

I can provide a code snippet if needed

lyio:
@StephenOTT  doesn't the API simply return an empty array and not nil?  I just had to build this myself  in my script and did exactly what you described. It's not pretty at all.

StephenOTT:
@lyio Yes you are correct it is a empty array.  Wrote that post while walking down the street.... :)

Yes it is not Pretty for sure, hence defined as the "monkey wrench way"

StephenOTT:
The code i there together real quick looked something like this:

```ruby


issuePageNum = 1
issues = @glClient.issues(projectID, :per_page=>100, :page=>issuePageNum)
while issues.length > 0  do

	# Iterates through each issue and get the notes and merges the notes into the issue hash.
	issues.each do |x|
		x = x.to_h	# Converts the ObjectifiedHash that is returned by GitLab into a Ruby Hash

		commentPageNum = 1
		issueComments = @glClient.issue_notes(x["project_id"], x["id"], :per_page=>100, :page=>commentPageNum)	# Gets the notes for the current issue

		if issueComments.length > 0	# If there are notes in the issue then...

			while issueComments.length > 0  do  # Loop used to control api pagination
				issueComments.each do |y|	# For Each Comment...
					y = y.to_h	# Convert from ObjectifiableHash to Ruby Hash / Do what you want to do to each Comment

				# Part of the pagination process to go through as many pages of comments that may exist for the current issue
				commentPageNum += 1
				issueComments = @glClient.issue_notes(x["project_id"], x["id"], :per_page=>100, :page=>commentPageNum)	# Gets the notes for the current issue
			end
		end
	end

	issuePageNum += 1
	issues = @glClient.issues(projectID, :per_page=>100, :page=>issuePageNum)
end

```

NARKOZ:
#134

PR #100, opened 2015-01-07
Added support for repository files create, edit and remove.

razielgn:
The original doc is at http://doc.gitlab.com/ce/api/repository_files.html.
I didn't implement the GET as the functionality is already present, albeit in a different module. Let me know if it should be added anyway.
Cheers.


NARKOZ:
Thank you.

PR #103, opened 2015-01-22
added username parameter to user

vincenzodibiaggio:
with accordingly of the gitlab-api docs I have added the username parameter to the user endpoint

https://github.com/gitlabhq/gitlabhq/blob/master/doc/api/users.md


NARKOZ:
Thank you for your PR! 

However it's a breaking change. See #62 #89
I'll accept it if you implement it in non-breaking way.

vincenzodibiaggio:
Ok, look if it's good for you. 
Thanks

vincenzodibiaggio:
Also added delete method

NARKOZ:
I meant to add `username` param, but implement it a way that doesn't break existing behaviour. 

For example: check if the last argument is Hash or String.

lyio:
Also, as mentioned in #105 you can simply search by username with
```ruby
Gitlab.users(search: 'user_name')
```

NARKOZ:
Implemented in #133

PR #107, opened 2015-02-04
adds feature to create a config file

logicminds:
previously you were required to specify a configuration via
environment variables.  With this change you can now optionally
specify the same configs in a config file.
Additionally if you have multiple gitlab servers that you connect
to you can easily switch by specifying a default config file to use.
This is handy where the config files are named after servers.


NARKOZ:
I appreciate your pull request, but it introduces unnecessary complexity to the code. 

> you can easily switch by specifying a default config file to use

You can accomplish the same thing by placing the env variables in the file and sourcing it.

-
Rel: #78

PR #109, opened 2015-02-05
Authenticate via oauth an auth_token

### LINKED ISSUE #104: Update for GitLab oAuth support - oAuth2 Token
Current wrapper supports "private tokens" but does not appear to have support for the new oAuth2 Access Tokens as described here: http://doc.gitlab.com/ce/api/oauth2.html

dsander:
This PR allows to authenticate via an oauth auth_token, should fix #104

### DIFF
--- a/lib/gitlab/configuration.rb
+++ b/lib/gitlab/configuration.rb
@@ -2,7 +2,7 @@ module Gitlab
   # Defines constants and methods related to configuration.
   module Configuration
     # An array of valid keys in the options hash when configuring a Gitlab::API.
-    VALID_OPTIONS_KEYS = [:endpoint, :private_token, :user_agent, :sudo, :httparty].freeze
+    VALID_OPTIONS_KEYS = [:endpoint, :private_token, :auth_token, :user_agent, :sudo, :httparty].freeze
 
     # The user agent that will be sent to the API endpoint if none is set.
     DEFAULT_USER_AGENT = "Gitlab Ruby Gem #{Gitlab::VERSION}".freeze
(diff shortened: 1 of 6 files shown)


NARKOZ:
Does it work if you pass OAuth access token instead of `private_token`?

dsander:
I tried that, but passing the `auth_token` as `private_token` does not work on my 7.7 instance.

lib/gitlab/configuration.rb
```diff
@@ -32,6 +32,7 @@ def options
     def reset
       self.endpoint       = ENV['GITLAB_API_ENDPOINT']
       self.private_token  = ENV['GITLAB_API_PRIVATE_TOKEN']
+      self.auth_token     = ENV['GITLAB_API_AUTH_TOKEN']
```
NARKOZ: I think better to have `oauth = false` instead of introducing `auth_token`. So user can set only `private_token` and change `oauth` option if needed.

dsander: I was going with `auth_token` and `private_token` because it follows the gitlab wording, `private_token` for the personal tokens and `auth_token` for OAuth applications. You would rather use the `private_token` for both authentication schemes and have an `oauth` option to toggle between them?

NARKOZ: Yes, I want to avoid confusion with existing `private_token`.
Another option would be to check the length of supplied token and auto-detect.

dsander: For users which keep using their `private_token` nothing changes. I am sorry, I don't how adding a new key for a different authentication method could cause confusion?

NARKOZ: There is no case when user needs to set both `auth_token` and `private_token`. 
So, I propose to have only 1 config key (`private_token`), add alias (`oauth_token`) for it and auto detect correct header later behind the scenes. Does it make sense?

NARKOZ:
@dsander do you want to update it to what I suggested?

dsander:
Sorry, I was busy the last weeks. Does this look more like what you had in mind?

NARKOZ:
Sorry for the delay! Thank you.

StephenOTT:
:+1:

dsander:
Thanks for reviewing and merging it!

Issue #105, opened 2015-01-26
Support "search" api calls


StephenOTT:
The latest api for gitlab has many need for using search to find information.  such as: get projects/search/:query

StephenOTT:
would be nice if there was support for the various search features such as:

![screen shot 2015-01-25 at 8 58 58 pm](https://cloud.githubusercontent.com/assets/1994838/5894430/0979af4a-a4d5-11e4-99d4-69f33a7521dd.png)

asedge:
The `projects` method can be used for searching and supports all the same options:
```
2.2.0 :027 > Gitlab.projects(search: 'proj').map { |p| p.name }                                                                                                                                                           
 => ["proj1", "proj2"] 
2.2.0 :028 > Gitlab.projects(search: 'proj', order_by: 'name', sort: 'desc').map { |p| p.name }
 => ["proj2", "proj1"] 
2.2.0 :029 > Gitlab.projects(search: 'proj', order_by: 'name', sort: 'asc', per_page: 1).map { |p| p.name }
 => ["proj1"] 
```

asedge:
Reopen if this needs further discussion.  Thanks!

StephenOTT:
Thanks!

StephenOTT:
@asedge 
Can you show an example with Namespace/Project ?
I am getting varying results and just trying to hunt down the issue.

asedge:
@StephenOTT Looking at the definition of the search method (here: https://github.com/gitlabhq/gitlabhq/blob/master/app/models/project.rb#L200-L207) you probably won't be able to search like 'Namespace/Project'.  Just one or the other.  Once you get the results back you could always use the `reject` method to filter further.

StephenOTT:
@asedge is there a restriction on being able to search all of GitLab?  I seem to only be able to return projects that I am the owner of.  I cannot seem to do a search through all public repos or private repos that i can see... ?

Thanks

PR #112, opened 2015-02-25
Add project_search method

ey3ball:
There is currently no way to search for a project, add a project_search method based on gitlab's /project/search API

### DIFF
--- a/lib/gitlab/client/projects.rb
+++ b/lib/gitlab/client/projects.rb
@@ -19,6 +19,23 @@ class Gitlab::Client
       end
     end
 
+    # Search for projects by name
+    #
+    # @example
+    #   Gitlab.project_search('gitlab')
+    #   Gitlab.project_search('gitlab', :order_by => 'last_activity_at')
+    #
+    # @param  [Hash] options A customizable set of options.
+    # @option options [String] :per_page Number of projects to return per page
+    # @option options [String] :page The page to retrieve
+    # @option options [String] :order_by Return requests ordered by id, name, created_at or last_activity_at fields
+    # @option options [String] :sort Return requests sorted in asc or desc order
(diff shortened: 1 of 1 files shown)


asedge:
Thanks for this PR, but I don't think this method is necessary.  You can already do searches using the `projects` method.  Have a look at this closed issue for some more details: #105.  If I remember correctly it also searches more than just the project name (project path, namespace and description).

ey3ball:
This way of doing things does not work with older gitlab versions, for instance on my server (gitlab 7.6.2) :

$ gitlab projects "{search: 'test'}"
Will return all projects and does not filter.

Digging further it seems this facility was only added very recently in gitlab 7.7.0

See : https://github.com/gitlabhq/gitlabhq/commit/8689ce1efef8438debeec2a3a6d669f4d5a435c4 (Jan. 12, 2015 !)

Moreover, I think that whenever there is a REST API in gitlab it makes sense to have the corresponding command in this project for the sake of consistency / completeness.

What do you think ?

Thanks,

Florian

asedge:
I'll reopen it and ping @NARKOZ for comment.

It would need tests to be merged but you may want to wait to hear back first.

NARKOZ:
@asedge @ey3ball makes sense to add it.

@ey3ball could you also add tests?

ey3ball:
Ok, I'l look into yet and update the PR

StephenOTT:
:+1:  The `projects` command seems to be inconstant in its delivery of results.  Plus it seems to focus only on repos you own.

asedge:
I wrote some tests for this so I merged it.  Thanks!

StephenOTT:
GREAT!!! Thanks. This will make GitLab-Time-Tracking much easier to build :)

PR #117, opened 2015-03-06
"Commits API" functions added, also added edit_project

marc-ta:


lib/gitlab/client/commits.rb
```diff
@@ -0,0 +1,84 @@
+class Gitlab::Client
+  # Defines methods related to repositories.
```
NARKOZ: Please add `@see https://github.com/gitlabhq/gitlabhq/blob/master/doc/api/commits.md` below it.

NARKOZ:
@marc-ta thank you for your contributions.

Please don't change file modes to executable (chmod -x), and squash your commits.

NARKOZ:
Could you leave only commit 6765929fea0f9335a54a0c0a572753750b7b9d82? Because commits API are already covered in [client/repositories.rb](https://github.com/NARKOZ/gitlab/blob/master/lib/gitlab/client/repositories.rb).

asedge:
@marc-ta The "commits api" methods already exist in [repositories.rb](https://github.com/NARKOZ/gitlab/blob/master/lib/gitlab/client/repositories.rb).  Splitting them out into a new file is good but can you please remove the old method definitions, tests, and fixtures so we can avoid duplication?  Thanks!

NARKOZ:
Thank you, merged manually 257737c80ca93a71dfb5f8d990e5de1423603dfc

PR #118, opened 2015-03-06
Add create_merge_request_note

onionjake:
Add `create_merge_request_note` to correspond to this API call:

https://github.com/gitlabhq/gitlabhq/blob/master/doc/api/notes.md#create-new-merge-request-note

### DIFF
--- a/spec/gitlab/client/notes_spec.rb
+++ b/spec/gitlab/client/notes_spec.rb
@@ -152,5 +152,22 @@ describe Gitlab::Client do
         expect(@note.author.name).to eq("John Smith")
       end
     end
+
+    context "when merge_request note" do
+      before do
+        stub_post("/projects/3/merge_requests/7/notes", "note")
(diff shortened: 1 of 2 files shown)


onionjake:
I'm not really sure why the tests failed?  As far as I can tell I am matching the existing tests.

spec/gitlab/client/notes_spec.rb
```diff
@@ -152,5 +152,22 @@
         expect(@note.author.name).to eq("John Smith")
       end
     end
+
+    context "when merge_request note" do
+      before do
+        stub_post("/projects/3/merge_request/7/notes", "note")
```
asedge: I think the test failed because you stubbed `merge_request` in the path but it really should be `merge_requests`.

onionjake: Good catch.  I also was expecting a post to 'snippets'.

onionjake:
It looks fixed.

asedge:
Thanks!

Issue #114, opened 2015-03-03
Get issues for milestone


pbendersky:
GitLab has an API point to get all issues from a given milestone.

`GET /projects/:id/milestones/:milestone_id/issues`

I want to send a PR with this addition, but wanted to discuss first what's the best place to implement it without breaking the library's API. The `issues` method in https://github.com/NARKOZ/gitlab/blob/master/lib/gitlab/client/issues.rb#L17-L23 receives a project parameter, but not a milestone.

Should I add a `milestone_issues` method that gets a project and milestone parameter?

NARKOZ:
> Should I add a `milestone_issues` method that gets a project and milestone parameter?

Yes, the method should be added to https://github.com/NARKOZ/gitlab/blob/master/lib/gitlab/client/milestones.rb

asedge:
@pbendersky If your commit message includes "closes|fixes #<issue_id>" it will automatically close the issue when we merge it.  Just in case you didn't know. Thanks again! :+1:

pbendersky:
Forgot about it, sorry. :)

PR #119, opened 2015-03-06
Added support to get milestone issues.

pbendersky:
Added support (and test) for API endpoint to get a single milestone's issues, as reported in #114


asedge:
Looks good, thanks for the code!

PR #132, opened 2015-05-15
Added Users#delete_user method

cthulhu666:
I needed a way to delete users, and it was not supported, so I've added it.


asedge:
Thanks!

PR #134, opened 2015-06-17
Another implementation of utilizing pagination headers

nanofi:
I've attempted to implement #51 in a different way of #85. Instead of returning all response as an array, my implementation returns a wrapper class of array response; this has the accessors of first/last/next/previous pages. `.auto_pagination` method enables to iterate all items.

Examples:
```ruby
# An array response
@projects = Gitlab.projects(per_page: 5)
# Checks existence of the next page
@projects.has_next_page
# Retrieves the next page
@projects.next_page
# Iterates all projects
@projects.auto_paginate do |project|
  # do something
end
# Retrieves all projects as an array
@projects.auto_paginate
```


lib/gitlab/array_response.rb
```diff
@@ -0,0 +1,90 @@
+module Gitlab
+  # Wrapper class of array response.
+  class ArrayResponse
```
NARKOZ: Let's name it as `PaginatedResponse`

lib/gitlab/array_response.rb
```diff
@@ -0,0 +1,90 @@
...
+      @client.get(path)
+    end
+    def has_prev_page
+      !(@links.nil? or @links.prev.nil?)
+    end
+    def prev_page
+      return nil if @client.nil? or !has_prev_page
+      path = @links.prev.sub(/#{@client.endpoint}/, '')
+      @client.get(path)
+    end
```
NARKOZ: Please add empty lines between method definitions

NARKOZ:
I like this approach.

NARKOZ:
Thank you.

PR #137, opened 2015-07-03
deleting a gitlab project returns the string "true"

tosmi:
JSON.parse converts "true" to a boolean, but the custom parsing function
just checks if the body is an array or hash.

So check if the body is a real boolean and if it's true, return an
empty hash.

This is with gitlab 7.11.4, didn't check any other versions...

### DIFF
--- a/lib/gitlab/request.rb
+++ b/lib/gitlab/request.rb
@@ -19,6 +19,10 @@ module Gitlab
         ObjectifiedHash.new body
       elsif body.is_a? Array
         body.collect! { |e| ObjectifiedHash.new(e) }
+      elsif body
+        true
+      elsif !body
+        false
       elsif body.nil?
         false
       else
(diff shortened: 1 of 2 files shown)


lib/gitlab/request.rb
```diff
@@ -19,6 +19,8 @@ def self.parse(body)
         ObjectifiedHash.new body
       elsif body.is_a? Array
         body.collect! { |e| ObjectifiedHash.new(e) }
+      elsif !!body == body
+        ObjectifiedHash.new(Hash.new)
```
NARKOZ: Let's return `true` when response body is "true".

tosmi:
thanks for the hint, i wasn't quite sure what parse should return in this case. i just looked at the documentation for the delete_project method... this should be fixed now.

lib/gitlab/request.rb
```diff
@@ -19,6 +19,8 @@ def self.parse(body)
         ObjectifiedHash.new body
       elsif body.is_a? Array
         body.collect! { |e| ObjectifiedHash.new(e) }
+      elsif !!body == body
```
NARKOZ: This statement looks odd. Can't we just write `body == 'true'`?

tosmi:
ok, if changed the check and added some more tests.

tosmi:
and finally if the body contains 'false' return the boolean false.

NARKOZ:
Thanks. Can you squash the commits please?

tosmi:
and squashed they are

thanks
toni

NARKOZ:
@tosmi thank you!

PR #139, opened 2015-07-13
Added support description option in create_group method

azomazo:
In the description of the method ```create_group``` indicated that the parameter ```description``` supported, but the method signature does not support it.

### DIFF
--- a/spec/gitlab/client/groups_spec.rb
+++ b/spec/gitlab/client/groups_spec.rb
@@ -21,19 +21,40 @@ describe Gitlab::Client do
   end
 
   describe ".create_group" do
-    before do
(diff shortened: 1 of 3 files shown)


spec/gitlab/client/groups_spec.rb
```diff
@@ -37,6 +37,25 @@
     end
   end
 
+  describe ".create_group_with_description" do
```
asedge: Can you make this a context under the already existing `describe ".create_group"` block?

azomazo:
I moved the code under the context. That's better?

asedge:
@azomazo Looks good, thanks!

PR #142, opened 2015-07-18
Add a `--json` parameter to the CLI commands

michaellihs:
Adding a `--json` parameter to a CLI command returns the output in JSON. Example:

````
gitlab project 605 --json

>>>> {
       "cmd": "Gitlab.project 605",
       "result": {
         "archived": false,
         "avatar_url": null,
         "created_at": "2015-05-21T13:50:16.514Z",
         "default_branch": "master",
         "description": "",
        ...
       }
     }
````
For "single row" returned by the command, the `result` field will be a map. If there are multiple rows returned, the `result`field will be an array of maps.

You are then able to do things like this (given you have [jq](http://stedolan.github.io/jq/) installed on your system):

````
gitlab project PROJECT_ID --json | jq '.result.id'

>>>> PROJECT_ID
````

The JSON is pretty-printed to make it readable for humans :-)

### DIFF
--- a/lib/gitlab/cli_helpers.rb
+++ b/lib/gitlab/cli_helpers.rb
@@ -1,4 +1,6 @@
 require 'yaml'
+require 'json'
+
 
 class Gitlab::CLI
   # Defines methods related to CLI output and formatting.
@@ -99,16 +101,29 @@ class Gitlab::CLI
       end
     end
 
+    def output_json(cmd, args, data)
+      if data.empty?
+        puts '{}'
+      else
+        hash_result = case data
+          when Gitlab::ObjectifiedHash
+            record_hash([data], cmd, args, true)
+          when Array
+            record_hash(data, cmd, args)
+          else
+            { :cmd => cmd, :data => data, :args => args }
+        end
+        puts JSON.pretty_generate(hash_result)
+      end
+    end
+
     # Table to display records.
     #
     # @return [Terminal::Table]
     def record_table(data, cmd, args)
       return 'No data' if data.empty?
 
-      arr = data.map(&:to_h)
-      keys = arr.first.keys.sort {|x, y| x.to_s <=> y.to_s }
-      keys = keys & required_fields(args) if required_fields(args).any?
-      keys = keys - excluded_fields(args)
+      arr, keys = get_keys(args, data)
 
       table do |t|
         t.title = "Gitlab.#{cmd} #{args.join(', ')}"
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Comment by joachimmathes** [comment]
+1 for this pull request

**Comment by dgrammlich** [comment]
+1!

File: lib/gitlab/cli_helpers.rb [hunk 1]
```diff
@@ -99,16 +101,29 @@ def output_table(cmd, args, data)
       end
     end
 
+    def output_json(cmd, args, data)
+      if data.empty?
+        puts '{}'
+      else
+        case data
```
asedge: `hash_result = case data` is preferable here instead of repeating the assignment.

lib/gitlab/cli_helpers.rb
```diff
@@ -134,6 +149,54 @@ def record_table(data, cmd, args)
...
+    # @param  [Array]  args         Options passed to the API call
+    # @param  [bool]   single_value If set to true, a single result should be returned
+    # @return [Hash]   Result hash
+    def record_hash(data, cmd, args, single_value = false)
+      if data.empty?
+        result = nil
+      else
+        arr, keys = get_keys(args, data)
+        result = []
+        arr.each_with_index do |hash, index|
```
asedge: Unless I've missed it, you don't actually use the index.  You should probably just use `each` instead.

spec/gitlab/cli_spec.rb
```diff
@@ -59,6 +60,21 @@
...
+
+    context "when command with json output" do
+      before do
+        stub_get("/user", "user")
+        args = ['user', '--json']
+        @output = capture_output { Gitlab::CLI.start(args) }
+      end
+
+      it "should render output as json" do
+        expect(JSON.parse(@output)['result']).to eq(JSON.parse(File.read(File.dirname(__FILE__) + '/../fixtures/user.json')))
```
asedge: I'm not an expert at writing tests, so don't assume I know what I'm talking about here, but I think maybe you should check here that some of the individual attributes (email, id, etc.) are correct instead.

michaellihs:
hey @asedge - thx for your feedback, I will try to fix your issues tomorrow! Just would be nice to know whether there are any plans to actually merge this, since I've been waiting for almost 2 months now... @NARKOZ - any feedback for me?

asedge:
@michaellihs I can merge it.  I just figured @NARKOZ is probably busy because it's been a while since you submitted this PR.  If you'd like to wait for @NARKOZ to review it I have no issue with that. :smile:

spec/gitlab/cli_spec.rb
michaellihs: Hi @asedge - I investigated that and from my understanding this is what happens in the test. A User is faked via the stub and we check that the JSON formatted output has the expected fields. So the attributes like email etc. are checked "implicitly".

michaellihs:
Hi @asedge - I refactored the code again to reflect your comments besides the last one, because I think the test is already doing exactly what you mentioned. If you disagree, please let me know and I will give it a try again. Otherwise, it would be great if you could merge it since we work with a patched version at the moment and it would be great to switch back to the official gem... thanks a lot for your help! Michael

michaellihs:
Hey @asedge - any news on this one? Would be great, if you could merge it!

asedge:
@michaellihs Please revert `006bf6c56058664d3645f31c1a21492cd5994d93` and changes to `CHANEGELOG.md`.  @NARKOZ will update those when he does a new release. Once you do that, please squash commits and we should be OK to merge.

michaellihs:
@asedge @NARKOZ - I did as you said and squashed my commits... hope it will be merged soon...

asedge:
Thanks for your contribution!

michaellihs:
@asedge Thanks for your help!

PR #145, opened 2015-08-07
Add Unprocessable error handler

ondra-m:
For example when you create a merge request with the same source and target branch you got_

_You can not use same project/branch for source and target_

and status 422.


asedge:
Thanks!

PR #146, opened 2015-08-19
Update README.md

raine:


asedge:
You can also set `GITLAB_API_HTTPARTY_OPTIONS` as a YAML hash, so for example: 
```export GITLAB_API_HTTPARTY_OPTIONS="{verify: false}"```
This would disable SSL verification in HTTParty - in case you were using a self-signed certificate.  Would you mind adding that to your example as well?

raine:
added

asedge:
Thanks!

PR #147, opened 2015-08-23
Adding sudo option for when forking a project

gregoriomelo:
This adds the option to fork a project for another user. One use case for such feature is when you want to fork projects for new users as soon as their account is created.

Using `Gitlab.sudo = 'someone_else'` might not be suitable in this case, because concurrent requests with different values for `sudo` can have a negative impact on each other.

Note: `g = Gitlab.client; g.sudo = 'someone_else'` didn't have the expected effect. This PR is my attempt to overcome it.


lib/gitlab/client/projects.rb
```diff
@@ -326,11 +326,15 @@ def delete_deploy_key(project, id)
...
+    #   Gitlab.create_fork(42, :sudo => 'another_username')
     #
     # @param  [Integer] project The ID of a project.
+    # @param  [Hash] options A customizable set of options.
+    # @option options [String] :sudo The username the project will be forked for
     # @return [Gitlab::ObjectifiedHash] Information about the forked project.
-    def create_fork(id)
-        post("/projects/fork/#{id}")
+    def create_fork(id, options = {})
+      body = {}.merge(options)
```
asedge: Lets skip the merge here.  Just pass `options` directly as the `body` and if we need to do a merge later we can do it.

gregoriomelo:
Thanks, @asedge. How does it look like now?

Cheers!

asedge:
@gregoriomelo Looks good, please squash the commits.

gregoriomelo:
Cool, @asedge. Squashed.

Cheers!

asedge:
Thanks!

PR #153, opened 2015-10-23
Add commit status API, was added in Gitlab 8.1

dsander:


asedge:
Could you rename the `commit_statuses` method to `commit_status`?  This would keep it consistent with other method naming.  Otherwise, looks good!

dsander:
Sure, done.

asedge:
Thanks for the contribution!

PR #155, opened 2015-11-02
Adding an all_users method

Karamorf:
The all_users method wraps the paging of the users call into an Enumerator.

I got tired of writing the paging logic over and over whenever I needed to iterate over all of the users. This method solves that for me without immediately requesting every user in GitLab.


NARKOZ:
Pagination methods were added to the master and will be available with next release.

PR #164, opened 2015-12-15
Update README.md

walterheck:
Add some information on how to use this gem as a cli.

### DIFF
--- a/README.md
+++ b/README.md
@@ -105,8 +105,7 @@ For more information, refer to [documentation](http://rubydoc.info/gems/gitlab/f
 
 ## CLI
 
-Configuration example:
-
+It is possible to use this gem as a command line interface to gitlab. In order to make that work you need to set a few environment variables:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -116,16 +115,23 @@ export GITLAB_API_PRIVATE_TOKEN=<your private token from /profile/account>
 export GITLAB_API_HTTPARTY_OPTIONS="{verify: false}"
 ```
 
+Usage:
+
+When you want to know which CLI commands are supported, take a look at the client [commands implemented in this gem](http://www.rubydoc.info/gems/gitlab/3.4.0/Gitlab/Client). Any of those methods can be called as a command by passing the parameters of the commands as parameters of the CLI.
```
NARKOZ: You can get a list of available commands via CLI help. 

`gitlab help` lists available topics. `gitlab help Topic` will list available actions.

PR #166, opened 2016-01-02
Added possibility to change emails of user

azomazo:


lib/gitlab/client/emails.rb
```diff
@@ -0,0 +1,60 @@
+class Gitlab::Client
+  # Defines methods related to user emails.
+  # @see https://github.com/gitlabhq/gitlabhq/blob/master/doc/api/users.md
+  module Emails
```
NARKOZ: Put methods into existing Users module.

lib/gitlab/client/emails.rb
```diff
@@ -0,0 +1,60 @@
...
+    # @example
+    #   Gitlab.delete_email(2)
+    #   Gitlab.delete_email(3, 2)
+    #
+    # @param  [Integer] id Email address ID
+    # @param  [Integer] user_id The ID of a user.
+    # @return [Boolean]
+    def delete_email(id, user_id=nil)
+      url = user_id.to_i.zero? ? "/user/emails/#{id}" : "/users/#{user_id}/emails/#{id}"
+      delete(url, parser: ->(body, _) { body == 'null' ? true :  Gitlab::Request.parse(body) })
```
NARKOZ: Parsing should be handled in [Request class](https://github.com/NARKOZ/gitlab/blob/master/lib%2Fgitlab%2Frequest.rb#L15-L15)

azomazo:
Moved code Users and remove override parser in `delete_email` method.
But method `delete_email` always returns `null`. Therefore, the result of the method turns out false, which can be misleading.

NARKOZ:
Thank you.

>Therefore, the result of the method turns out false, which can be misleading.

Yes, but that's because of the GitLab API's inconsistency.

PR #171, opened 2016-01-25
Use `respond_to_missing?` instead of `respond_to?`

tsigo:
See https://robots.thoughtbot.com/always-define-respond-to-missing-when-overriding

Prior, I was unable to mock the client, for example by doing:

```ruby
expect(Gitlab).to receive(:protect_branch).with(anything(), anything())
```

### DIFF
--- a/lib/gitlab.rb
+++ b/lib/gitlab.rb
@@ -25,8 +25,8 @@ module Gitlab
   end
 
   # Delegate to Gitlab::Client
-  def self.respond_to?(method)
-    client.respond_to?(method) || super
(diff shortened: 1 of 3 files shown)


tsigo:
Would appreciate a minor version bump and release after merging. I need this for some work we're doing on [release-tools](https://gitlab.com/gitlab-org/release-tools). Thanks!

NARKOZ:
Can you change in 

```
lib/gitlab/objectified_hash.rb
lib/gitlab/paginated_response.rb
``` 

too? Thanks.

tsigo:
@NARKOZ Good call, updated.

PR #172, opened 2016-01-27
New builds endpoints

kmarcisz:
With Gitlab 8.4 new endpoint has been introduced for builds. This pull has the method calls and tests coverage for the new functionality.


NARKOZ:
Thanks.

PR #177, opened 2016-02-03
CLI all results and CLI YAML options

dldinternet:
A paginated response will only display first page worth of results ...
Command line args like 'endpoint: https://gitlab..../' is ignored and results in "Please set an endpoint to API"


asedge:
Sorry to close this on you.  If you're still interested in this open a new issue and we can discuss a way to implement it.  Thanks!

PR #180, opened 2016-02-14
Add API to get a list of projects under a group

shadeslayer:
This allows the user to get a list of projects under a specific group.

### DIFF
--- a/lib/gitlab/client/groups.rb
+++ b/lib/gitlab/client/groups.rb
@@ -120,5 +120,15 @@ class Gitlab::Client
       options[:search] = search
       get("/groups", query: options)
     end
+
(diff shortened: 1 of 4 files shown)


shadeslayer:
A quick note about this, since I forgot to mention it last night, apparently this will only fetch the first 20 results on the page. So perhaps we should add a option argument to get more pages? Though the API docs don't document this, so someone will have to check if that actually works.

shadeslayer:
Apparently this is not really required since specifying a group_id parameter in the Gitlab.projects call seems to work just fine now. I guess the Gitlab instance was just slow perhaps since it was not returning anything earlier.

PR #186, opened 2016-03-26
Update YARD so it can be formatted easier for the CLI.

asedge:
Ping @NARKOZ.  I wasn't sure if you would like this but it definitely would improve the `help` output in the CLI since it looks a bit like the YAML strings needed to pass `options`.


asedge:
Merged and pushed.

Issue #162, opened 2015-12-13
Viewing and Assigning a user to an issue in the cli (currently displayed as "Hash")


e72575f3:
Thanks for fixing the other bug so fast!!

I'm trying to figure out how to view and assign an assignee to an issue in the cli, currently the output for Assignee and Author is `Hash`, can you give an example on how to change the assignee? Thanks!
```
+----------+--------+--------------------------+-------------+--------+-----+--------+-----------+------------+--------+--------------------+--------------------------+
| assignee | author | created_at               | description | id     | iid | labels | milestone | project_id | state  | title              | updated_at               |
+----------+--------+--------------------------+-------------+--------+-----+--------+-----------+------------+--------+--------------------+--------------------------+
| Hash     | Hash   | 2015-12-09T20:04:21.293Z |             | 012345 | 12  | []     | null      | 654321     | opened | delete some folder | 2013-11-10T13:27:24.553Z |
+----------+--------+--------------------------+-------------+--------+-----+--------+-----------+------------+--------+--------------------+--------------------------+
```

asedge:
You may be better off using the `--json` option for now, which should display the full `Hash` for both `assignee` and `author`.

I think we can improve the output though and use the `id` field, if it exists, here instead of the literal `Hash`: https://github.com/NARKOZ/gitlab/blob/master/lib/gitlab/cli_helpers.rb#L137

I replaced that line on my local copy with this:
```value = value.has_key?('id') ? value['id'] : 'Hash'```

Seemed to work and is definitely more useful than `Hash`.

PR #189, opened 2016-03-31
Improve record table output to use Hash's 'id' field if it exists. Fi…

asedge:
…xes #162.


asedge:
Merged this in already.

PR #196, opened 2016-04-26
Add `run_trigger` API

edgemaster:
The API to execute a trigger is not documented on the same page as the
rest of the trigger management API, but this seems the most appropriate
place for it.

The `trigger` method has been renamed to `trigger_detail` to prevent
confusion as to its behaviour, it retrieves details about a trigger, and
doesn't cause a trigger to be activated.


edgemaster:
Actually, after a second look at this patch, I think I've implemented it and the tests incorrectly. Will revise it.

lib/gitlab/client/build_triggers.rb
```diff
@@ -21,7 +22,7 @@ def triggers(project)
     # @param  [Integer] project The ID of a project.
     # @param  [String] token The token of a trigger.
     # @return [Gitlab::ObjectifiedHash] The trigger.
-    def trigger(project, token)
+    def trigger_detail(project, token)
```
asedge: I think `trigger` was a fine name for this method.  It matches the pattern used elsewhere in the gem where asking for `project` or `user` gets you details.   What do you think?

edgemaster: The difference here is that `trigger` could be interpreted as either a verb or a noun, it was this ambiguity I was attempting to avoid.
Do you have any views on the naming of the counterpart `trigger_build` method?

asedge: I hear what you are saying.  I guess all things considered it made sense to me to match the pattern of other Gitlab objects.  Plus the inline doc you provided states its purpose:
```
# Gets details of project's build trigger.
```
`trigger_build` seems like a good name to me.

edgemaster: Ok, will revert back and update the PR at some point.

asedge: OK, ping me when you do.

edgemaster:
I'm working on rebasing this to the v4 API.

edgemaster:
I have rebased this PR onto the patch in #275, and updated it to be compatible with API v4.

asedge:
@edgemaster I think I'll just close #275 and merge this one.  If I merge #275 I'm pretty sure there will be a conflict because there will be a merge commit between the two commits listed here.

asedge:
Oh neat, it auto closed #275 when I merged this PR.

PR #202, opened 2016-05-30
Add user_search method

Dreeg:
Inspired by group_search method
https://github.com/NARKOZ/gitlab/commit/ac595af15a07601b3b657d4b4c1c479651aaa7d7


asedge:
Thank you for your contribution and for, probably unintentionally, pointing out that I shouldn't have created that method since the functionality already exists through options passed to the `groups` method.  Any reason you can't pass `:search` to the `users` method to achieve the same result?

Dreeg:
Hi asedge,
the first reason is for completeness, as for the groups, also for the users a method well documented and available.
Moreover, my fault, I did not try to pass 'options' undocumented. Indeed, as also for the groups, it works.

At this point my edit is right for convenience (as always for the groups).

I take this opportunity to thank for the project and for the help!

asedge:
@Dreeg  I was just saying that I probably shouldn't have done https://github.com/NARKOZ/gitlab/commit/ac595af15a07601b3b657d4b4c1c479651aaa7d7 since the functionality already existed.  I don't want us to get in the habit of creating a bunch of convenience methods when we could pass `options` to already existing methods instead.

Dreeg:
I didn't understand, I'm sorry.
In this case you can see what to do :)

Thanks again for everything!

asedge:
Sorry for the delay.  I decided to merge this.  Thanks for the contribution!

PR #204, opened 2016-06-01
Add build_artifacts method

nanofi:
This PR adds a method of downloading artifacts.  `build_artifacts.json` in fixtures is a zip file containing a text file `example.txt`.


asedge:
Thanks!  Sorry for the delay.

It would be nice if this was usable from the CLI too.  Not sure how that would work.

nanofi:
I found a bug due to missing the error handling. Fixed it, and added a test for it. 

I add some code to work the method in the CLI. Here are sample outputs. 
table output:
```
+------+---------------+
| Gitlab.build_artifacts 8, 3 |
+------+---------------+
| data | filename      |
+------+---------------+
| File | artifacts.zip |
+------+---------------+
```
json output:
```
{
  "cmd": "Gitlab.build_artifacts 8, 3",
  "result": {
    "data": "<base64 encoded file content>",
    "filename": "artifacts.zip"
  }
}
```

asedge:
Thanks!

PR #210, opened 2016-07-26
Implemented tags API methods

### LINKED ISSUE #199: Feature requests: add tags support
Please correct me if I am wrong, but is the tag part of the gitlab API not currently supported in this client?

https://github.com/gitlabhq/gitlabhq/blob/master/doc/api/tag.md

If that's the case, I'd be happy to do the work to put them in.

jblackman:
This resolves #199.

Implemented tags API methods. Moved the existing tags() and create_tag() methods from repository.rb to tags.rb.

Sorry about the long delay between raising the FR and implementing it.


lib/gitlab/client/tags.rb
```diff
@@ -0,0 +1,96 @@
...
+    #
+    # @example
+    #   Gitlab.tag(3, 'api')
+    #   Gitlab.repo_tag(5, 'master')
+    #
+    # @param  [Integer] project The ID of a project.
+    # @param  [String] tag The name of the tag.
+    # @return [Gitlab::ObjectifiedHash]
+    def tag(project, tag)
+      get("/projects/#{project}/repository/tags/#{tag}", query: {})
```
asedge: Can't we just leave `query: {}` off here?

asedge:
Looks pretty good, just one comment that could use addressing.  Thanks!

lib/gitlab/client/tags.rb
jblackman: Shall do

asedge:
Thank you!

PR #211, opened 2016-08-03
Add in GitlabCI Runner support

davidcollom:
This pull request (as titled) adds in runner support:

* List users runners
* List all runners (if admin)
* List projects runners (specific)
* Get runner
* Update runner
* Delete runner
* Enable project runner
* Disable project runner

### DIFF
--- a/lib/gitlab/client/runners.rb
+++ b/lib/gitlab/client/runners.rb
@@ -1,0 +1,115 @@
+class Gitlab::Client
+  # Defines methods related to runners.
+  # @see https://github.com/gitlabhq/gitlabhq/blob/master/doc/api/runners.md
+  module Runners
+
+    # Get a list of specific runners available to the user.
+    # Full runner params documentation: https://github.com/gitlabhq/gitlabhq/blob/master/doc/api/runners.md
+    #
+    # @example
+    #   Gitlab.runners
+    #   Gitlab.runners(:active)
+    #   Gitlab.runners(:paused)
+    #
+    # @param  [Hash] options A customizable set of options.
+    # @option options [String] :scope The scope of specific runners to show, one of: active, paused, online; showing all runners if none provided
+    # @return [Array<Gitlab::ObjectifiedHash>]
+    def runners(options = {})
+      get("/runners", query: options)
+    end
+
+    # Get a list of all runners in the GitLab instance (specific and shared). Access is restricted to users with admin privileges.
+    # Full runner params documentation: https://github.com/gitlabhq/gitlabhq/blob/master/doc/api/runners.md
+    #
+    # @example
+    #   Gitlab.all_runners
+    #
+    # @param  [Hash] options A customizable set of options.
+    # @option options [String] :scope The scope of runners to show, one of: specific, shared, active, paused, online; showing all runners if none provided
+    # @return [Array<Gitlab::ObjectifiedHash>]
+    def all_runners(options = {})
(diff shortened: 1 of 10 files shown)


lib/gitlab/client/runners.rb
```diff
@@ -0,0 +1,124 @@
...
+    #
+    # @example
+    #   Gitlab.project_disable_runner(2, 42)
+    #
+    # @param  [Integer, String] id The ID of a project.
+    # @param  [Integer, String] id The ID of a runner.
+    # @return <Gitlab::ObjectifiedHash>
+    def project_disable_runner(project_id, id)
+      body = { runner_id: id }
+      delete("/projects/#{project_id}/runners/#{id}", body: body)
```
asedge: Request body is not needed to disable a runner according to this doc: https://github.com/gitlabhq/gitlabhq/blob/master/doc/api/runners.md.  Have you observed something different?

lib/gitlab/client/runners.rb
```diff
@@ -0,0 +1,124 @@
...
+    #
+    # @param  [Hash] options A customizable set of options.
+    # @option options [String] :scope The scope of specific runners to show, one of: active, paused, online; showing all runners if none provided
+    # @return [Array<Gitlab::ObjectifiedHash>]
+    def runners(options = {})
+      if options[:scope]
+        get("/runners", query: options)
+      else
+        get("/runners")
+      end
```
asedge: Do we need the check here?  Couldn't we just pass the empty hash down regardless with no ill effect?

davidcollom: I was following the convention from https://github.com/NARKOZ/gitlab/blob/master/lib/gitlab/client/projects.rb#L15

```
    def projects(options={})
      if options[:scope]
        get("/projects/#{options[:scope]}", query: options)
      else
        get("/projects", query: options)
      end
    end
```
but I guess not.

lib/gitlab/client/runners.rb
davidcollom: Agree'd I've left this in from the enable - will remove.

lib/gitlab/client/runners.rb
asedge: Fair enough. :)

asedge:
Thanks for the PR!  What is ` spec/fixtures/runners_edit.json` used for?  It's an empty file that does not appear to be referenced anywhere in the code.

davidcollom:
`spec/fixtures/runners_edit.json` was added by mistake, I decided to use `spec/fixtures/runner.json` seen as though the response was technically no different.

lib/gitlab/client/runners.rb
```diff
@@ -0,0 +1,119 @@
...
+    #
+    # @param  [Hash] options A customizable set of options.
+    # @option options [String] :scope The scope of runners to show, one of: specific, shared, active, paused, online; showing all runners if none provided
+    # @return [Array<Gitlab::ObjectifiedHash>]
+    def all_runners(options = {})
+      if options[:scope]
+        get("/runners/all", query: options)
+      else
+        get("/runners/all")
+      end
```
asedge: Can you change this also so it's like the `runners` method?

asedge:
@davidcollom One more thing to fix up and then you can go ahead and squash commits.

asedge:
Thanks for the contribution!

Issue #217, opened 2016-09-04
Update JSON gem 


connorshea:
As mentioned in https://github.com/danger/danger/issues/521 could we update the gem to use JSON 2.x? Or even just drop the dependency entirely since most functionality is built into Ruby nowadays.

k0nserv:
To my understanding that gem is only being included for Ruby versions less than 2.0 per https://github.com/NARKOZ/gitlab/blob/master/gitlab.gemspec#L22-L24 It's included via HTTParty

connorshea:
Ah, I guess it would be worth considering dropping support for anything less than Ruby 2.1 in that case. But this isn't the issue to discuss that in :) Closing.

orta:
This isn't strictly true - just checked, as with CocoaPods, Rubygems parses the ruby file and then submits that up to the Gem server as a compiled JSON package: https://rubygems.org/api/v1/gems/gitlab.json

As this gem was submitted with a 1.9 version of ruby, then everyone who uses Rubygems from the main API gets HTTParty 0.13

```json
"runtime": [
  {
    "name": "httparty",
    "requirements": "~> 0.13.0"
  },
  {
    "name": "terminal-table",
    "requirements": ">= 0"
  }
]
```

In theory [this line in the Gemspec](https://github.com/NARKOZ/gitlab/blob/master/gitlab.gemspec#L22) would need to be 

```ruby
  if RUBY_VERSION < '2.0' && ENV['TESTING']
```

Or something similar, to ensure the tests still pass, as that's why it was added: https://github.com/NARKOZ/gitlab/commit/268a19b7d4f92a66fd17e50e09b73601abdb20c0

asedge:
@dblessing and @jvanbaarsen -  Wanted to see if you had any feedback on dropping support for Ruby 1.9.3 for version 4.0.

connorshea:
@asedge I don't have any problem with it, the GitLab app doesn't support 1.9.3 and the most major project (that I know of) dependent on this gem is Danger, which doesn't support 1.9.3 either.

jvanbaarsen:
@asedge Let's do this :-)

connorshea:
This can be closed now :)

PR #237, opened 2016-09-16
Lock terminal-table version to provide support for Ruby 1.9 and 2.0.

connorshea:
1.9.x and 2.0.x are both EOL, they're no longer supported with security releases. Travis builds were failing due to terminal-table releasing a breaking change for old Ruby versions in a patch version.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ language: ruby
 rvm:
   - 1.9.3
   - 2.0
-  - 2.1
-  - 2.2
-  - 2.3.0
+  - 2.1.9
+  - 2.2.5
+  - 2.3.1
 before_install: gem update bundler
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ language: ruby
 rvm:
   - 1.9.3
   - 2.0
-  - 2.1.9
-  - 2.2.5
-  - 2.3.1
+  - 2.1
+  - 2.2
+  - 2.3
 before_install: gem update bundler
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 ## Change Log
 
+### Unreleased
+
+- Lock terminal-table to prevent build failures on Ruby 1.9/2.0. (@connorshea)
+
 ### 3.7.0 (16/08/2016)
 
 - Add in GitlabCI Runner support (@davidcollum)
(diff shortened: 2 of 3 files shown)


jvanbaarsen:
This would mean however that we have to make a major version bump (To 4.0)

connorshea:
This is because terminal-table introduced a breaking change: https://github.com/tj/terminal-table/issues/79

Ruby 2.0 is EOL, but Danger still supports it, so I suppose we should remain compatible.

connorshea:
@jvanbaarsen we can remain compatible by just locking the terminal-table version, though I don't see a huge amount of use in continuing to support 1.9.

connorshea:
@jvanbaarsen 1.9 support retained, builds are passing again :)

asedge:
> This would mean however that we have to make a major version bump (To 4.0)

I've been considering this based on #217.  I would be OK dropping support for 1.9.x in 4.0 but maybe not 2.0.  Thoughts?

connorshea:
@asedge I'd be fine with that :)

connorshea:
I would say that should be separate, however.

connorshea:
Separate from this PR, that is. (Can't edit comments on GitHub mobile ಠ_ಠ)

connorshea:
@jvanbaarsen @asedge can this be merged so tests will pass for other PRs?

.travis.yml
```diff
@@ -2,7 +2,7 @@ language: ruby
 rvm:
   - 1.9.3
   - 2.0
-  - 2.1
-  - 2.2
-  - 2.3.0
```
asedge: I think we need to just drop the bugfix/patch digit for `2.3.0` to use the latest version.  See the build [here](https://travis-ci.org/NARKOZ/gitlab/jobs/160698118) it uses `2.2.5` not `2.2.0`.   The Travis docs also indicate that `2.1`, `2.2`, etc., are aliases: https://docs.travis-ci.com/user/languages/ruby#Choosing-Ruby-versions-and-implementations-to-test-against.

asedge:
Hm, this review feature is weird.

connorshea:
@asedge I hear GitLab is pretty good these days ;)

connorshea:
Do I need to rebase for the changes to nullify the review? This is weird.

asedge:
Apparently not.  Thanks!

PR #240, opened 2016-09-19
Move share_project_with_group entry to Unreleased section

connorshea:
Simple Changelog change.


asedge:
Thanks, I was going to move it but this saves me the effort.

PR #242, opened 2016-09-20
Allow to retrieve ssh_keys for a specific user

dirker:
This pull request implements the API call for https://docs.gitlab.com/ce/api/users.html#list-ssh-keys-for-user.

### DIFF
--- a/lib/gitlab/client/users.rb
+++ b/lib/gitlab/client/users.rb
@@ -127,13 +127,20 @@ class Gitlab::Client
     #
     # @example
     #   Gitlab.ssh_keys
+    #   Gitlab.ssh_keys({ user_id: 2 })
     #
     # @param  [Hash] options A customizable set of options.
     # @option options [Integer] :page The page number.
     # @option options [Integer] :per_page The number of results per page.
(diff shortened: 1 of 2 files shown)


lib/gitlab/client/users.rb
```diff
@@ -127,13 +127,18 @@ def session(email, password)
...
     #   Gitlab.ssh_keys
+    #   Gitlab.ssh_keys(2)
     #
     # @param  [Hash] options A customizable set of options.
     # @option options [Integer] :page The page number.
     # @option options [Integer] :per_page The number of results per page.
     # @return [Array<Gitlab::ObjectifiedHash>]
-    def ssh_keys(options={})
-      get("/user/keys", query: options)
+    def ssh_keys(user_id=nil, options={})
```
asedge: I'm not fond of changing the method signature.  What if we look for `:user_id` as part of options and call the different route if that key exists instead?

jvanbaarsen: 👍  For making it part of the option.

Changing the method signature is not an option without deprecation.

dirker:
Reworked, now the user_id can be specified via the options.

asedge:
Thanks!

PR #243, opened 2016-09-24
Remove the Ruby 1.x support in the gemspec

orta:
This is the result from the discussion in https://github.com/NARKOZ/gitlab/issues/217 👍

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,5 @@
 language: ruby
 rvm:
-  - 1.9.3
   - 2.0
   - 2.1
   - 2.2
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ### Unreleased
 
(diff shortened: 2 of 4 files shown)


connorshea:
LGTM 👍

jvanbaarsen:
Maybe we should also include this requirement in the README somewhere? (To prevent bug reports for users on older rubies)

orta:
> Might want to add this to the Changelog though.

hah, I know just the tool for that ;)

orta:
👍 done both

asedge:
@orta Squash commits please.

orta:
I shall, but if that's something you care about, you can do this as a part of the merge button rather than asking contributors to do it:

screen shot 2016-09-27 at 10 13 52

orta:
👍 done

asedge:
Thanks, I suppose I forgot that existed.

PR #245, opened 2016-10-10
Mask the token when inspect the client object

JuanitoFatas:
This PR masks the token when invoking `inspect` on gitlab client.

So when people do bug report, they won't accidentally pasted token information. [other client library](https://github.com/octokit/octokit.rb/blob/c1e66093eeb499c59e49212ca10c2a4efc356836/lib/octokit/client.rb#L118-L133) also does this.


asedge:
Thanks!

PR #247, opened 2016-10-26
Add methods to disable and enable deploy keys on projects.

buzzdeee:
While I was looking how to best transfer git repos from bitbucket to gitlab, I found this gem very useful.
I have to transfer a lot of repos, on the first repo, I have to create deploy keys, but then later on, I only needed to enable those already created deploy keys on the other repos.

I added two functions: enable_deploy_key and disable_deploy_key to the client/repositories.rb
file, to accompany the other deploy keys functions.
These parts of the API are described here:
https://docs.gitlab.com/ee/api/deploy_keys.html#enable-a-deploy-key

Further added spec tests for these new functions.

### DIFF
--- a/lib/gitlab/client/projects.rb
+++ b/lib/gitlab/client/projects.rb
@@ -364,6 +364,30 @@ class Gitlab::Client
       post("/projects/#{project}/keys", body: { title: title, key: key })
     end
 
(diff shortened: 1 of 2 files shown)


asedge:
Thank you for your contribution!

PR #249, opened 2016-11-02
Protect & unprotect single branch api supported.

leopku:
First of all, thanks for this amazing project.

Protect & unprotect single branch api supported.

The api was here

https://docs.gitlab.com/ce/api/projects.html#protect-single-branch

and here

https://docs.gitlab.com/ce/api/projects.html#unprotect-single-branch

### DIFF
--- a/lib/gitlab/client/projects.rb
+++ b/lib/gitlab/client/projects.rb
@@ -443,5 +443,32 @@ class Gitlab::Client
     def unstar_project(id)
       delete("/projects/#{id}/star")
     end
+
+    # Protect single branch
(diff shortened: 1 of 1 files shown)


asedge:
Thank you for the contribution however we already have coverage for these methods, see here: https://github.com/NARKOZ/gitlab/blob/master/lib/gitlab/client/branches.rb#L34-L60

In the future you should have a look at the [CONTRIBUTING.md](https://github.com/NARKOZ/gitlab/blob/master/CONTRIBUTING.md) for tips on how to effectively contribute to this project.

Thanks again!

PR #251, opened 2016-12-01
CLI Output record as list and simple key value format

bergholdt:
Option: --list
For data with many keys a table output is not a god match,
this change makes it possible to output data in list format

Option --simple
This format output each record as lines of key=value, if you
only fetch one field it will only print value.
The goal is to have a format that's simple to parse

### DIFF
--- a/lib/gitlab/cli.rb
+++ b/lib/gitlab/cli.rb
@@ -46,10 +46,7 @@ class Gitlab::CLI
     when 'shell'
       Gitlab::Shell.start
     else
-      if args.include? '--json'
-        @json_output = true
-        args.delete '--json'
-      end
+      @output_format = extract_output_format(args)
 
       unless valid_command?(cmd)
         puts "Unknown command. Run `gitlab help` for a list of available commands."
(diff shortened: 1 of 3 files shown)


criztovyl:
Whats the state of this PR?
I would really like to see this feature in the CLI. :)

asedge:
@criztovyl I honestly don't see why it's needed.  I've been waiting to see if someone would make a case for it. We already have a `--json` option which is easy to parse.  If you use the `jq` command line tool you can select the bits of information you need.  What does this provide that cannot be achieved with `--json`?

bergholdt:
My use case for creating the PR was that 
- I was on a Windows Enterprise environment and had build agents running Batch. 
- I was restricted in what tools could be install
- I wanted the to keep the code that developers needed to write in gitlab.yml simple.

However, I'm no longer on this Windows environment :-) 
I don't know if my old team are still using my fork and have need for this PR.

criztovyl:
@asedge
I agree.
My intention only was to express that I like (not need) this feature, á la nice to have. :)

timofonic:
@bergholdt Maybe others like @criztovyl may want this feature, anyway :)

@AAraKKe @NARKOZ @cvand @asedge Any hopes to have this pull request merged or reimplemented?

asedge:
Sorry everyone, I just don't think I'm convinced this is necessary.  Thanks for the PR @bergholdt and sorry it did not end up getting merged.  I do appreciate your efforts!

PR #253, opened 2016-12-08
Feature/230 note that project id can also be a string

### LINKED ISSUE #230: Note that project ID can also be a string
You can pass the project name as a string rather than the ID as an integer to access a project via the API, e.g. `https://gitlab.com/api/v3/projects/gitlab-org%2Fgitlab-ce/issues?private_token=PRIVATE_TOKEN`. This makes everything a lot easier to work with, also note the slash must be encoded for this to work. 

A lot of the gem's docs have to be updated with this.

See also https://gitlab.com/gitlab-org/gitlab-ce/merge_requests/6102 and https://gitlab.com/gitlab-org/gitlab-ce/merge_requests/6133.

bergholdt:
Make it possible to use NAMESPACE/PROJECT_NAME on all methods that uses project ID

ex.
  gitlab project gitlab-org/gitlab-ce
  gitlab milestones gitlab-org/gitlab-ce
  gitlab tree gitlab-org/gitlab-ce
  gitlab tags gitlab-org/gitlab-ce
  ...

Closes #230

### DIFF
--- a/lib/gitlab/client.rb
+++ b/lib/gitlab/client.rb
@@ -39,6 +39,10 @@ module Gitlab
       inspected
     end
 
+    def url_encode(s)
+      ERB::Util.url_encode(s)
+    end
+
     private
 
     def only_show_last_four_chars(token)
--- a/lib/gitlab/client/branches.rb
+++ b/lib/gitlab/client/branches.rb
@@ -7,13 +7,13 @@ class Gitlab::Client
     # @example
     #   Gitlab.branches(42)
     #
-    # @param  [Integer] project The ID of a project.
+    # @param  [Integer, String] project The ID or name of a project.
     # @param  [Hash] options A customizable set of options.
     # @option options [Integer] :page The page number.
     # @option options [Integer] :per_page The number of results per page.
     # @return [Array<Gitlab::ObjectifiedHash>]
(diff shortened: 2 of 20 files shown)


agimenez:
It is pretty needed :)

asedge:
This is already possible if you url encode your project string before using it as a parameter to one of the methods.

I am not fond of adding code where documentation would do.  @jvanbaarsen - I know you're busy but do you have an opinion here?

bergholdt:
Egen using CLI we should not expect users to url encode

asedge:
Good point.  Have you written any tests that verify this functionality? I don't remember seeing any.

bergholdt:
@asedge I have added specs to verify that project string is url_encoded.

Also made code change to support backword compatibility for users that already have url_encoded.

bergholdt:
Note to self - always run all specs locally before push.

Removed the backward compatibility code again...as it was not that compatible 😊

asedge:
Thanks!

orta:
Any chance we can get a gem release with this in?

asedge:
@orta Yes, "soon".  Trying to button a few things up first.

asedge:
@orta `4.0.0` has been released.

PR #255, opened 2017-01-30
Allow issues to use NAMESPACE/REPO identifier

brodock:
You can use `int` based ID to identify a project, or you can use `NAMESPACE/REPO` according to the API documentation: https://docs.gitlab.com/ce/api/projects.html#get-single-project

The advantage of using `NAMESPACE/REPO` everywhere, is that you don't need to first load the repository to get the numeric ID so you can use it on other places.

The gem itself already allows the use of the "literal" identification, but there is one check in the issues getter that prevents it's use. This PR fixes that.

cc @NARKOZ @connorshea @asedge

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 - Update documentation to link to docs.gitlab.com instead of the GitHub mirror for GitLab CE. (@connorshea)
 - Add method `share_project_with_group` (@danhalligan)
 - Allow to retrieve `ssh_keys` for a specific user(@dirker)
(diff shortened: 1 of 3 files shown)


asedge:
Thanks for the contribution but I think this was already covered in #253!

bergholdt:
@asedge we still need this one. 
Cifrene Issue code was checking for integer value

bergholdt:
Cifrene = current 
iPhone autocorrect was on wrong language

asedge:
OK, merged.  Thanks!

PR #268, opened 2017-03-30
Add appropriate Content-Type header

mltsy:
This makes it easier to tests using a rack-based Gitlab mock API. (Sinatra in my case)

Rack will assume POST requests include `www-form-urlencoded` data, but it won't assume the same of PUT or PATCH requests, and therefore won't parse the body unless a header is present.  Granted, this header is obviously not required for Gitlab to accept the request, but it's the right way to construct the request in general, which should make it more reusable/compatible with testing frameworks, etc.

It seems like `HTTParty` or even `Net::HTTP` is making the decision to encode the data as form-encoded data, so why it doesn't automatically attach that header, I'm not sure.  But that's further than I have time to investigate/fix at the moment :)

Note: I have not yet tested this patch with an actual Gitlab instance, but if all the requests are sent as form-encoded data (and as far as I know they all are, or the data is included in the query-string), it shouldn't cause any problems.

### DIFF
--- a/lib/gitlab/request.rb
+++ b/lib/gitlab/request.rb
@@ -6,7 +6,7 @@ module Gitlab
   class Request
     include HTTParty
     format :json
-    headers 'Accept' => 'application/json'
+    headers 'Accept' => 'application/json', 'Content-Type' => 'application/x-www-form-urlencoded'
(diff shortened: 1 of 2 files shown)


asedge:
Thank you!

PR #273, opened 2017-04-24
Remove version-lock for terminal-table

SuperTux88:
The latest version of `terminal-table` is compatible with ruby 2.0 again, see https://github.com/tj/terminal-table/pull/80


asedge:
Thanks!

PR #274, opened 2017-04-27
client: add support for the keys resource

dirker:
This pull request adds support for the keys resource: https://docs.gitlab.com/ce/api/keys.html.


asedge:
Thanks!

PR #278, opened 2017-05-19
add jobs endpoint methods

hjanuschka:
adds the https://docs.gitlab.com/ee/api/jobs.html#jobs-api

i had the requirement to play a manual job, and discovered that the gem does not support the job api. this PR adds the api to the gem.

adds:
  * Gitlab.jobs
  * Gitlab.pipline_jobs
  * Gitlab.job
  * Gitlab.job_artifacts
  * Gitlab.job_artifacts_download
  * Gitlab.job_trace
  * Gitlab.job_retry
  * Gitlab.job_cancel
  * Gitlab.job_erase
  * Gitlab.job_play
  * Gitlab.job_artifacts_keep


lib/gitlab/client/jobs.rb
```diff
@@ -0,0 +1,158 @@
...
+  module Jobs
+    # Gets a list of Jobs for a Project
+    #
+    # @example
+    #   Gitlab.jobs(1)
+    #   Gitlab.jobs("project")
+    #
+    # @param  [Integer, String] id The ID or name of a project.
+    # @return [Array<Gitlab::ObjectifiedHash>]
+    def jobs(project_id, scope='created')
```
asedge: Scope is not actually a required field. Why set a default that's different from the default defined in the API doc?  The default scope would be to get all jobs.

I think it would be better if you followed the pattern here: https://github.com/NARKOZ/gitlab/blob/master/lib/gitlab/client/merge_requests.rb#L16-L17

Then if more options get added in the future there's not much to change.

lib/gitlab/client/jobs.rb
```diff
@@ -0,0 +1,158 @@
...
+    # Gets a list of Jobs from a pipeline
+    #
+    # @example
+    #   Gitlab.pipeline_jobs(1, 2)
+    #   Gitlab.pipeline_jobs("project", 2)
+    #
+    # @param  [Integer, String] The ID or name of a project.
+    # @param  [Integer]  the id of the pipeline
+    # @return [Array<Gitlab::ObjectifiedHash>]
+    def pipeline_jobs(project_id, pipeline_id)
```
asedge: This can also take the scope, optionally, so do the same as you do for your `jobs` method.

lib/gitlab/client/jobs.rb
```diff
@@ -0,0 +1,158 @@
...
+  # Defines methods related to projects.
+  # @see https://docs.gitlab.com/ee/api/jobs.html
+  module Jobs
+    # Gets a list of Jobs for a Project
+    #
+    # @example
+    #   Gitlab.jobs(1)
+    #   Gitlab.jobs("project")
+    #
+    # @param  [Integer, String] id The ID or name of a project.
```
asedge: You are missing the 2nd parameter here.

asedge:
@hjanuschka Need to update the tests.  Mind adding cases for no scope, 1 string scope and an array of strings scope?

hjanuschka:
tests fixed, and scope spec's added.

asedge:
Thanks!

PR #279, opened 2017-05-30
Rename branch_name parameter to branch in create_branch method

sr189:
In Gitlab 9.0 the `branch_name` parameter for the `create_branch` method was renamed to `branch` (see: https://gitlab.com/gitlab-org/gitlab-ce/issues/22132)

This pull request renames the parameter for the client.


asedge:
Thanks!

PR #281, opened 2017-06-02
Modify and delete wall notes, issue notes, snippet notes and MR notes

akkee:
I was using the Gitlab API to use issue logging in one of my applications. Couldn't find wrappers for editing and deleting issue comments. So just trying to contribute here.

So we already have `modify_merge_request_note` and `delete_merge_request_note`, inside merge_requests API, but not for wall notes, issue notes and snippet notes. Why not move all of the notes related work to Notes API

MR contains the following new Client API methods:
1. **delete_note** - Deletes a wall note
2. **delete_issue_note** - Deletes an issue note
3. **delete_snippet_note** - Deletes a snippet note
4. **delete_merge_request_note** - Deletes a merge request note
5. **edit_note** - Modifies a wall note
6. **edit_issue_note** - Modifies an issue note
7. **edit_snippet_note** - Modifies a snippet note
8. **edit_merge_request_note** - Modifies a merge request note

### DIFF
--- a/lib/gitlab/client/merge_requests.rb
+++ b/lib/gitlab/client/merge_requests.rb
@@ -82,65 +82,6 @@ class Gitlab::Client
       put("/projects/#{url_encode project}/merge_requests/#{id}/merge", body: options)
     end
 
-    # Adds a comment to a merge request.
-    #
-    # @example
-    #   Gitlab.create_merge_request_comment(5, 1, "Awesome merge!")
-    #   Gitlab.create_merge_request_comment('gitlab', 1, "Awesome merge!")
-    #
-    # @param  [Integer, String] project The ID or name of a project.
-    # @param  [Integer] id The ID of a merge request.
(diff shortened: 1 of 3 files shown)


asedge:
This looks pretty good but we have some overlap of methods: https://github.com/NARKOZ/gitlab/blob/master/lib/gitlab/client/merge_requests.rb#L85-L143

Would you mind cleaning some of that up as well?  Ideally everything related to leaving comments/notes should be in `notes.rb`.  We should probably provide aliases for methods we had previously defined so we're not making an API change in what will be a minor release.  Does that make sense and do you mind adding (or removing) those things to your PR?

akkee:
That makes perfect sense, and ideally what I said too. Made the changes. Have a look and see if we can go forward with this.

akkee:
Rebased commits force-closed this PR, opened #283 for the same

PR #286, opened 2017-06-21
Add Environments module

### LINKED ISSUE #285: Add `Environments` API endpoints
There is a (new?) set of endpoints for `environments` in Gitlab:

https://docs.gitlab.com/ce/api/environments.html

Add methods to correspond to these endpoints.

mltsy:
(to access new environments endpoints)

Fixes #285

### DIFF
--- a/lib/gitlab/client/environments.rb
+++ b/lib/gitlab/client/environments.rb
@@ -33,7 +33,7 @@ class Gitlab::Client
     #
     # @examples
     #   Gitlab.create_environment(5, 'test-branch')
-    #   Gitlab.create_environment(5, 'test-branch', 'https://test-branch.example.host.com')
+    #   Gitlab.create_environment(5, 'test-branch', external_url: 'https://test-branch.example.host.com')
     #
     # @param  [Integer, String] project The ID or name of a project.
     # @param  [String] env_name Name for the environment
--- a/lib/gitlab/client/environments.rb
+++ b/lib/gitlab/client/environments.rb
@@ -32,16 +32,15 @@ class Gitlab::Client
     # Create an environment.
     #
     # @examples
-    #   Gitlab.create_pipeline(5, 'test-branch')
-    #   Gitlab.create_pipeline(5, 'test-branch', 'https://test-branch.example.host.com')
+    #   Gitlab.create_environment(5, 'test-branch')
+    #   Gitlab.create_environment(5, 'test-branch', 'https://test-branch.example.host.com')
     #
     # @param  [Integer, String] project The ID or name of a project.
     # @param  [String] env_name Name for the environment
-    # @param  [String] external_url Optional URL for viewing the deployed project in this environment
+    # @option options [String] :external_url Optional URL for viewing the deployed project in this environment
     # @return [Gitlab::ObjectifiedHash] The updated environment.
-    def create_environment(project, env_name, external_url=nil)
-      body = {name: env_name}
-      body[:external_url] = external_url if external_url
+    def create_environment(project, env_name, options = {})
+      body = {name: env_name}.merge(options)
       post("/projects/#{url_encode project}/environments", body: body)
     end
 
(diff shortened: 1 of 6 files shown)


spec/gitlab/client/groups_spec.rb
```diff
@@ -59,20 +59,18 @@
   end
 
   describe ".delete_group" do
-    context "without description" do
```
mltsy: Pretty sure this context was a copy/paste typo, so I removed it.

mltsy:
Any chance of getting this merged soon @asedge?  Or is there anything else I can do to get it more ready to merge?

lib/gitlab/client/environments.rb
```diff
@@ -0,0 +1,88 @@
...
+    #
+    # @examples
+    #   Gitlab.create_pipeline(5, 'test-branch')
+    #   Gitlab.create_pipeline(5, 'test-branch', 'https://test-branch.example.host.com')
+    #
+    # @param  [Integer, String] project The ID or name of a project.
+    # @param  [String] env_name Name for the environment
+    # @param  [String] external_url Optional URL for viewing the deployed project in this environment
+    # @return [Gitlab::ObjectifiedHash] The updated environment.
+    def create_environment(project, env_name, external_url=nil)
```
asedge: I would recommend we use an `options` (default empty) hash as the 3rd argument.  Since `external_url` is optional and more optional parameters _could_ be added in the future this should reduce the need to change the method signature.

lib/gitlab/client/environments.rb
```diff
@@ -0,0 +1,88 @@
...
+      body = {name: env_name}
+      body[:external_url] = external_url if external_url
+      post("/projects/#{url_encode project}/environments", body: body)
+    end
+
+    # Update an environment.
+    #
+    # @examples
+    #   Gitlab.edit_environment(5, 36, name: 'test-branch')
+    #   Gitlab.create_pipeline(5, 36, external_url: 'https://test-branch.example.host.com')
```
asedge: Copy/paste error.

lib/gitlab/client/environments.rb
```diff
@@ -0,0 +1,88 @@
...
+    # @return [Gitlab::ObjectifiedHash]
+    def environment(project, id)
+      get("/projects/#{url_encode project}/environments/#{id}")
+    end
+
+    # Create an environment.
+    #
+    # @examples
+    #   Gitlab.create_pipeline(5, 'test-branch')
+    #   Gitlab.create_pipeline(5, 'test-branch', 'https://test-branch.example.host.com')
```
asedge: Copy/paste error

mltsy:
Yup, yup, good calls :wink:  All fixed!

lib/gitlab/client/environments.rb
```diff
@@ -32,24 +32,23 @@ def environment(project, id)
     # Create an environment.
     #
     # @examples
-    #   Gitlab.create_pipeline(5, 'test-branch')
-    #   Gitlab.create_pipeline(5, 'test-branch', 'https://test-branch.example.host.com')
+    #   Gitlab.create_environment(5, 'test-branch')
+    #   Gitlab.create_environment(5, 'test-branch', 'https://test-branch.example.host.com')
```
asedge: 3rd param of the example should be a hash now, right?

mltsy: Oops!  Yes, yes it should.  (I'm not very used to using YARD, clearly!)

asedge:
@mltsy Thanks!  In the future keep each PR to one topic.  I appreciate you fixing group_spec but filing a separate PR would be preferable.

mltsy:
Will do!  Yeah, I realized that was probably not the best idea after I'd pushed it (but by then it seemed like more work than it was worth to move it into another PR) - I'll try to be better about separating incidental fixes! :)

PR #287, opened 2017-06-22
Send create commit data via post body

sr189:
Currently when creating a commit the POST data is send via the query parameter. This leads to the Gitlab API to respond with 414 URI too long once there are several documents part of the commit.

This can be fixed when sending the data via the POST body attribute.


sr189:
Got a different commit in this pull request. Will create a new one.

PR #288, opened 2017-06-22
Use body parameter to send post data

sr189:
Currently when creating a commit the POST data is send via the query parameter. This leads to the Gitlab API to respond with 414 URI too long once there are several documents part of the commit.

This can be fixed when sending the data via the POST body attribute.


asedge:
Thanks for fixing this!

PR #291, opened 2017-06-30
Fix and clarify edit_project's option docs

mltsy:
Fixes a few copy and paste errors, along with clarifying that any valid option can be passed in.

### DIFF
--- a/lib/gitlab/client/projects.rb
+++ b/lib/gitlab/client/projects.rb
@@ -76,6 +76,7 @@ class Gitlab::Client
     # @param  [Hash] options A customizable set of options.
     # @option options [String] :description The description of a project.
     # @option options [String] :default_branch The default branch of a project.
+    # @option options [String] :path URL-safe repository name for new project. (Default is lowercase name with dashes)
     # @option options [String] :namespace_id The namespace in which to create a project.
     # @option options [Boolean] :wiki_enabled The wiki integration for a project (0 = false, 1 = true).
     # @option options [Boolean] :wall_enabled The wall functionality for a project (0 = false, 1 = true).
@@ -423,7 +424,7 @@ class Gitlab::Client
     # @param  [Integer, String] project The ID or name of a project.
     # @param  [Hash] options A customizable set of options
     # @option options [String] :name The name of a project
-    # @option options [String] :path The URL-safe path of the project (used in Gitlab URLs)
+    # @option options [String] :path The project's repository name, also used in Gitlab's URL path (letters, digits, `_`, `-` and `.` only)
     # @option options [String] :description The description to show in Gitlab
     # (Any provided options will be passed to Gitlab. See {https://docs.gitlab.com/ce/api/projects.html#edit-project Gitlab docs} for all valid options)
     #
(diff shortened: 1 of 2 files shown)


lib/gitlab/client/projects.rb
```diff
@@ -421,10 +421,12 @@ def create_fork(id, options={})
     #   Gitlab.edit_project(42, { name: 'project_name' })
     #
     # @param  [Integer, String] project The ID or name of a project.
-    # @param  [Hash] options A customizable set of options.
+    # @param  [Hash] options A customizable set of options
     # @option options [String] :name The name of a project
-    # @option options [String] :path The name of a project
-    # @option options [String] :description The name of a project
+    # @option options [String] :path The URL-safe path of the project (used in Gitlab URLs)
```
asedge: Since we have a `url_encode` method should we just encode the `:path` for them if it's been provided?

mltsy: Oh, good idea! :+1:

mltsy: Actually, on second thought... it would probably be safer to just pass the user-provided value through unchanged, to ensure they're getting the path they wanted (or an error if it's not valid).  If they want an automatic option, the path is auto-generated from the name, so they don't need to pass a path at all.  I'll update the docs to reflect that.

mltsy:
I'm not sure if this is better or just the same ;)  But either way, after considering it, I think if the user is passing in a path explicitly, we shouldn't have the gem mangle it at all.

asedge:
I didn't even look at the method the first time I reviewed this so I have some new thoughts.

URL encoding the `path` makes no sense because it's not part of the URL.  However, we should use `url_encode` on the `id` parameter as you can see in the doc: https://docs.gitlab.com/ce/api/projects.html#edit-project.

What makes you think `path` parameter has to be URL encoded?  Nothing in the doc seems to indicate that.

mltsy:
Oh, I *don't* think the path needs to be URL encoded, I just knew that it needed to be URL-safe, meaning basically it can't have slashes or spaces in it (because it is used as a URL path segment in Gitlab), but I guess it *is* more accurate and clear to just say what characters are allowed (letters, digits, '-', '_', and '.' according to Gitlab's interface)

Also, good point about URL-encoding the id.  I had't even really looked at updating the method itself, since it was working fine for me.  Do you want me to add that to this PR?

asedge:
Yes please add it to this PR.  My point about URL safe/encoded or whatever is I'm not sure we need to be saying that.  I'm hopeful that Gitlab will return a good error message if a user passes along an invalid `path` when trying to edit a project.  I suppose it doesn't matter either way.

mltsy:
There we go!  I'm not sure what Gitlab's API error messages look like for this call - it was actually kind of hard for me to find the allowed characters for the `path` field originally (I had to enter an invalid one in a running gitlab instance and see the error message there), so I thought it might be a helpful reference, but I did remove them for now since it was inconsistent with the create_project docs.  I'd be happy to add it back in in both places if you think it would be helpful though :)

asedge:
Nice, thanks!

PR #292, opened 2017-06-30
Update Branch docs and add options param to protect_branch

mltsy:
to handle `developers_can_push` and `developers_can_merge` as described in Gitlab docs:
https://docs.gitlab.com/ce/api/branches.html#protect-repository-branch

### DIFF
--- a/lib/gitlab/client/branches.rb
+++ b/lib/gitlab/client/branches.rb
@@ -36,12 +36,18 @@ class Gitlab::Client
     # @example
     #   Gitlab.protect_branch(3, 'api')
     #   Gitlab.repo_protect_branch(5, 'master')
+    #   Gitlab.protect_branch(5, 'api', developers_can_push: true)
+    #
+    # To update options, call `protect_branch` again with new options (i.e. `developers_can_push: false`)
     #
     # @param  [Integer, String] project The ID or name of a project.
     # @param  [String] branch The name of the branch.
-    # @return [Gitlab::ObjectifiedHash]
-    def protect_branch(project, branch)
-      put("/projects/#{url_encode project}/repository/branches/#{branch}/protect")
+    # @param  [Hash] options A customizable set of options.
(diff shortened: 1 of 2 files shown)


lib/gitlab/client/branches.rb
```diff
@@ -82,7 +88,6 @@ def create_branch(project, branch, ref)
     #
     # @param  [Integer, String] project The ID or name of a project.
     # @param  [String] branch The name of the branch to delete
-    # @return [Gitlab::ObjectifiedHash]
```
asedge: This does not return anything?

mltsy: It seems to always return `false` (Although I'm not sure where that comes from!)... so I figure it must be an incidental return value that is not of any real significance.  (The API documentation doesn't mention any specific return value either)

asedge: There's an example response in the document you linked.

mltsy: Really?  Hmm... I see example responses for all the endpoints except `DELETE`...  Ohh, maybe the git diff is confusing you?  This doc is for the `delete_branch` method, not `create_branch`... (?)

asedge: You're right, the offset in the diff looked like it was `create_branch`.  Whoops!  I still don't think it should be removed... Can we just say it returns a `Boolean`?

mltsy: Sure, I can if you want - but I think it always returns false... maybe there are times it would return true, but I don't know what they would be - do you?  It seems more confusing than helpful to document it without being able to give any description of what it represents. (All ruby functions return something, of course, but is it worth documenting if the return value is meaningless?)  I suppose I could document it as "meaningless return value" or something?

asedge: Alright, leave it out.  You've swayed me.

PR #293, opened 2017-06-30
Update repository files endpoint for APIv4

mltsy:
New path for all endpoints is `repository/files/:file_path` (instead of passing file_path as a parameter)

### DIFF
--- a/lib/gitlab/client/repository_files.rb
+++ b/lib/gitlab/client/repository_files.rb
@@ -52,7 +52,7 @@ class Gitlab::Client
     # @return [Gitlab::ObjectifiedHash]
     def create_file(project, path, branch, content, commit_message)
       post("/projects/#{url_encode project}/repository/files/#{url_encode path}", body: {
-        branch_name: branch,
+        branch: branch,
         commit_message: commit_message
       }.merge(encoded_content_attributes(content)))
     end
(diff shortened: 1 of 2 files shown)


lib/gitlab/client/repository_files.rb
```diff
@@ -51,9 +51,8 @@ def get_file(project, file_path, ref)
     # @param  [String] commit message.
     # @return [Gitlab::ObjectifiedHash]
     def create_file(project, path, branch, content, commit_message)
```
asedge: Looking at the docs there seem to be a bunch of optional parameters we don't currently support.  Do you think we should add `options={}` and update the YARD to include the optional parameters defined in the Gitlab API doc for `Repository Files`?

lib/gitlab/client/repository_files.rb
```diff
@@ -70,9 +69,8 @@ def create_file(project, path, branch, content, commit_message)
     # @param  [String] commit message.
     # @return [Gitlab::ObjectifiedHash]
     def edit_file(project, path, branch, content, commit_message)
```
asedge: Looking at the docs there seem to be a bunch of optional parameters we don't currently support. Do you think we should add options={} and update the YARD to include the optional parameters defined in the Gitlab API doc for Repository Files?

lib/gitlab/client/repository_files.rb
```diff
@@ -88,9 +86,8 @@ def edit_file(project, path, branch, content, commit_message)
     # @param  [String] commit message.
     # @return [Gitlab::ObjectifiedHash]
     def remove_file(project, path, branch, commit_message)
```
asedge: Looking at the docs there seem to be a bunch of optional parameters we don't currently support. Do you think we should add options={} and update the YARD to include the optional parameters defined in the Gitlab API doc for Repository Files?

mltsy:
That should do it! :)

asedge:
Thanks! 👍

PR #299, opened 2017-07-05
Add options to Labels#create_label

hlidotbe:
Labels can have a description and a priority in addition to name and color.
It's now possible to provide them when creating or editing a label.


asedge:
@hlidotbe Thanks!

PR #301, opened 2017-07-09
Replace `projects/search/{query}` by `projects?search={query}`

edaubert:
Current method `project_search` doesn't work.

Indeed, the APIv4 is now using query parameter to this.

I made the modifications according to this new behavior.

### DIFF
--- a/lib/gitlab/client/projects.rb
+++ b/lib/gitlab/client/projects.rb
@@ -34,7 +34,7 @@ class Gitlab::Client
     # @option options [String] :sort Return requests sorted in asc or desc order
     # @return [Array<Gitlab::ObjectifiedHash>]
     def project_search(query, options={})
-      get("/projects?search=#{query}", query: options)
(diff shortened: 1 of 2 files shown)


lib/gitlab/client/projects.rb
```diff
@@ -34,7 +34,7 @@ def projects(options={})
     # @option options [String] :sort Return requests sorted in asc or desc order
     # @return [Array<Gitlab::ObjectifiedHash>]
     def project_search(query, options={})
-      get("/projects/search/#{query}", query: options)
+      get("/projects?search=#{query}", query: options)
```
asedge: I think it would be better to do something like:
```
get("/projects", query: options.merge(search: query))
```

edaubert: Ok, I did it.

asedge:
Great!  Thank you!

PR #302, opened 2017-07-12
Use url_encode in all commit resources

grodowski:
For some reason two methods in `Commits` didn't use `url_encode` for id / name, which caused an error in my project. My suggestion is to keep it consistent.


asedge:
Thanks!

PR #306, opened 2017-07-19
Add `group_member` method (retrieve single member by ID)

mltsy:


mltsy:
I don't know why this was left out, but here it is! (pretty simple) :smile:

asedge:
Thanks! 👍

Issue #313, opened 2017-08-19
Please allow creating ssh keys for foreign user


formorer:
Hi,

please allow creating ssh keys for a different user. API allow that via:

`DELETE /users/:id/keys/:key_id`

PR #314, opened 2017-08-19
Allow adding of ssh keys for different users

formorer:
Hi, 

this is a fix for #313 that allows adding ssh keys for other users.

### DIFF
--- a/lib/gitlab/client/users.rb
+++ b/lib/gitlab/client/users.rb
@@ -270,7 +270,7 @@ class Gitlab::Client
     # @option options [String] :per_page Number of user to return per page
     # @option options [String] :page The page to retrieve
     # @return [Array<Gitlab::ObjectifiedHash>]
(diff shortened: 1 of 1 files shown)


asedge:
There are a couple problems here.
1) You've added a function (username_search) that's unrelated to the topic of this PR.  This is generally frowned upon, you really should file a separate PR for this function.
2) There are no tests added for new code paths.

timofonic:
@formorer Are you still interested in improving this pull request so it can get merged?

@AAraKKe  @NARKOZ @cvand @asedge What about this?

formorer:
I should look more often into my notifications, yes I am still interested we need that feature for Debians gitlab instance.

NARKOZ:
@formorer please fix the issues that @asedge had mentioned. I'll merge after that.

NARKOZ:
#413

PR #324, opened 2017-10-09
Add support for group level variables

hlidotbe:
Added support for group level variables as described here: https://docs.gitlab.com/ce/api/group_level_variables.html

Fixed a couple copy/paste errors in the documentation of variables and update_variable.


NARKOZ:
Thank you.

Issue #289, opened 2017-06-23
fork project does not url encode the project name


logicminds:
We need to url encode this project name if not already done so.
https://github.com/NARKOZ/gitlab/blob/master/lib/gitlab/client/projects.rb#L414

logicminds:
Looks like there are many methods that do not encode the name.  We should just do this in the request object automatically.  Thoughts?

asedge:
Ideally, yes we should.  I think it might end up being a rather large change to make it happen and I wasn't willing to invest the time.

logicminds:
I think we can escape much of this work if we test to see if the string is already encoded.

```ruby

def post(path, options={})
      set_httparty_config(options)
      set_authorization_header(options)
      path = url_encode(path) unless already_encoded?(path)
      validate self.class.post(@endpoint + path, options)
end

def already_encoded?(path)
  # this probably will not work but something along this lines 
  path.contains?('%2F') unless path.contains?(/\d/)
end
```

asedge:
You cannot URL encode the entire path.  You could probably `split(/\//)`, `map` each element with `url_encode`, then `join('/')`.  That's a crappy way to handle it but it might actually work.

NARKOZ:
Fixed in #325

PR #325, opened 2017-10-13
Error 404 not found when requesting branch, containing slash ('/'), from API

RunsFor:
Getting 404 error from gitlab when requesting branch from API:

```ruby
Gitlab.branch(project_id, branch_name)
```

It happen when `branch` variable contains slash in it.

I found that latest gitlab releases require to encode branch name in its request url.

Example:

1. 404 Not Found
  https://gitlab.com/api/v4/projects/14022/repository/branches/feature/gitaly-feature-flag

2. 200 OK (when logged in)
  https://gitlab.com/api/v4/projects/14022/repository/branches/feature%2Fgitaly-feature-flag

[Here is](https://gitlab.com/gitlab-org/gitlab-ce/issues/24471) related issue on gitlab project.
And one more [related issue](https://github.com/NARKOZ/gitlab/issues/289) on this project.

But even if it may be considered as a bug in gitlab, I think it's better to encode branch names when doing such api requests.

### DIFF
--- a/lib/gitlab/client/branches.rb
+++ b/lib/gitlab/client/branches.rb
@@ -27,7 +27,7 @@ class Gitlab::Client
     # @param  [String] branch The name of the branch.
     # @return [Gitlab::ObjectifiedHash]
     def branch(project, branch)
-      get("/projects/#{url_encode project}/repository/branches/#{branch}")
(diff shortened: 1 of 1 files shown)


NARKOZ:
Thank you.

Issue #331, opened 2017-10-28
Protected branches endpoint outdated


AAraKKe:
It seems that the protected branches endpoint being used is no longer valid.

The one being used is:

`/projects/#{url_encode project}/repository/branches/#{url_encode branch}/protect`

while the current one is

`/projects/#{url_encode project}/protected_branches`

providing the nam of the branch in the body. The same applies to unprotec branch. There the endpoint should be

`/projects/#{url_encode project}/protected_branches/#{url_enconde branch}`

I would love to modify it and make a pull request but tests would need to be changed and I would prefer not to mess with them if possible and more things might need to be changed.

Link to the documentation:
https://docs.gitlab.com/ce/api/protected_branches.html#protect-repository-branches

AAraKKe:
I needed this to be updated so I just created a pull request if you like to implement it.

PR #332, opened 2017-10-28
Implement new endpoint for protected branches

AAraKKe:
This is related to issue #331, I needed to fix it for a project so if you like feel free to take it.

I also modified the tests but I am not an expert on rspec, maybe there is something you would like to adapt on that side. I have tried this and seems to be working for my use case.

### DIFF
--- a/lib/gitlab/request.rb
+++ b/lib/gitlab/request.rb
@@ -32,7 +32,8 @@ module Gitlab
 
     # Decodes a JSON response into Ruby object.
     def self.decode(response)
-      JSON.load response
+      return JSON.load response if response
+      {}
     rescue JSON::ParserError
       raise Error::Parsing, 'The response is not a valid JSON'
     end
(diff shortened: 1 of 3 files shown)


AAraKKe:
I run the tests locally and they run, I am not sure why they are not running on travis.

AAraKKe:
I just noticed an issue, when the new delete method is called no response is sent and the decode method raises an issue, I am not sure the best way to fix it, if you let me know I could fix it before accepting the PR

lib/gitlab/request.rb
```diff
@@ -32,7 +32,8 @@ def self.parse(body)
 
     # Decodes a JSON response into Ruby object.
     def self.decode(response)
-      JSON.load response
+      return JSON.load response if response
+      {}
```
NARKOZ: Can you change this to `response ? JSON.load(response) : {}`?

AAraKKe: I just pushed this modification

NARKOZ:
Thank you.

AAraKKe:
No problem!

PR #341, opened 2017-11-29
Added subgroups and group_edit methods

Rubemlrm:
Hi there,

First of all thanks for the hard work, this week I've tried use the gem and i've found that are missing some api methods at groups.rb, so i take some time to add the following methods:

- List Groups Subgroups based on the group ID we send on the API call (https://docs.gitlab.com/ce/api/groups.html#list-a-groups-s-subgroups)
- Edit group method, based on the group ID we send on the API call (https://docs.gitlab.com/ce/api/groups.html#update-group)

Besides this, i've added the fixtures and specs for testing this new 2 methods.

### DIFF
--- a/lib/gitlab/client/groups.rb
+++ b/lib/gitlab/client/groups.rb
@@ -159,6 +159,13 @@ class Gitlab::Client
     #
     # @param [Integer] id the ID of a group
     # @param [Hash] options A customizable set of options.
+    # @option options [String] :skip_groups Skip the group IDs passed.
+    # @option options [String] :all_available Show all the groups you have access to (defaults to false for authenticated users).
+    # @option options [String] :search Return the list of authorized groups matching the search criteria.
+    # @option options [String] :order_by Order groups by name or path. Default is name.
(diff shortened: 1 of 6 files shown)


lib/gitlab/version.rb
```diff
@@ -1,3 +1,3 @@
 module Gitlab
-  VERSION = '4.2.0'.freeze
+  VERSION = '4.2.1'.freeze
```
NARKOZ: Please don't change this file

lib/gitlab/client/groups.rb
```diff
@@ -152,5 +152,42 @@ def group_search(search, options={})
     def group_projects(id, options={})
       get("/groups/#{id}/projects", query: options)
     end
+
+    # Get a list of subgroups under a group
+    # @example
+    #   Gitlab.group_projects(1)
```
NARKOZ: `Gitlab.group_subgroups(1)`

Rubemlrm:
hi there, sorry for the last reply. I've submitted the requested changes.

NARKOZ:
Thank you.

PR #347, opened 2018-01-18
Fixes error when tags contain special characters

cimnine:
When a git tag contain characters with a special meaning in a url, then this library would send requests to the wrong urls. Examples of such characters are '/' and '#'.

This PR makes sure that such characters are properly url-encoded.


NARKOZ:
Thank you.

Issue #304, opened 2017-07-16
Unable to update note 


sleekybadger:
Hi. I tried to update merge request comment via notes api and looks like current behavior not match [documentation](https://github.com/NARKOZ/gitlab/blob/master/lib/gitlab/client/notes.rb#L256):

```ruby
client = Gitlab.client(endpoint: 'https://gitlab.com/api/v4', private_token: 'token')
client.edit_merge_request_comment(project_id, merge_request_id, note_id, 'New body')

Gitlab::Error::BadRequest: Server responded with code 400, message: body is missing. Request URI: https://gitlab.com/api/v4/projects/project_id/merge_requests/merge_request_id/notes/note_id
```

However, I can update comment if i will pass body in hash: 

```ruby
client = Gitlab.client(endpoint: 'https://gitlab.com/api/v4', private_token: 'token')
client.edit_merge_request_comment(project_id, merge_request_id, note_id, { body: 'New body' })
```

This behavior same for other note update methods, but not for [create methods](https://github.com/NARKOZ/gitlab/blob/master/lib/gitlab/client/notes.rb#L121). So I wonder, should we change code to match documentation and other notes methods or change documentation?

asedge:
@sleekybadger If `body` is required when editing/updating a note, we should update the code so it follows the pattern in the create methods.  If it's optional we should update the documentation.  Does that make sense?  Will you be putting in a PR to fix this?

sleekybadger:
@asedge Yep, it's required. Sure, it's [here](https://github.com/NARKOZ/gitlab/pull/305)

asedge:
Oh crap.  I made a mistake.  I didn't even look at the behavior defined in the tests.  We cannot break the API as this will be a minor release.  You could make it so that we support both a String or a Hash with a `:body` key, but that could get a bit messy.  The other option is to just go back and update the documentation.  Sorry about this!

sleekybadger:
@asedge sure, pr updated

asedge:
@sleekybadger Sorry for letting this linger for so long.  I have merged #354 which should allow us to migrate to a consistent API while still warning users that passing the note content in a `Hash` is deprecated.  Thanks for bringing this to my attention and your attempts at fixing it - despite my lack of clear decision making!

PR #348, opened 2018-01-25
make edit_(.*_)?note accept the note body as a string again

evgeni:
the refactoring done in ceba47d6 made `edit_note`/`edit_*_note` expect the body in a Hash, instead if a plain String as it was before and as it is noted in the examples.

Refs: #304 
Yes, this breaks the API, but it was already broken for 4.2.0 by ceba47d6

### DIFF
--- a/lib/gitlab/client/notes.rb
+++ b/lib/gitlab/client/notes.rb
@@ -221,7 +221,7 @@ class Gitlab::Client
     # @param [Integer] id The ID of a note.
     # @return [Gitlab::ObjectifiedHash]
     def edit_note(project, id, body)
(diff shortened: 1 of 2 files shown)


asedge:
@NARKOZ I released some code that had an API that was inconsistent API other, related methods that already existed.  This PR makes them consistent.  I think we should merge and release this with an appropriate version that expresses the API may have changed, thoughts?

asedge:
I have handled this in #354 so we don't break backwards compatibility but can output a deprecation warning if people are passing a `Hash` for the body.  Thank you for your contribution!

evgeni:
thanks @asedge!

PR #352, opened 2018-01-31
Add featrues that allow users to specify namespace_id with strings

Sleeplessy:
I add a feature that allows users using String to describe their namespace_id while using command create_project. 
Something like `gitlab create_project 'gitlab' "{namespace_id : 'sleeplessy' }"`.
So that user can directly create repos under a group without trying to getting the **value of namespace_id**
And plz notify that the former help infomation of this command tells user that namespace_id should be a String,but [The Gitlab Official API DOC](https://docs.gitlab.com/ee/api/projects.html#create-project) tells that this param should be a Integer.
If the namespace_id is a Integer,or it wasn't defined,it would act like former.

### DIFF
--- a/lib/gitlab/client/projects.rb
+++ b/lib/gitlab/client/projects.rb
@@ -87,9 +87,25 @@ class Gitlab::Client
     # @option options [Integer] :user_id The user/owner id of a project.
     # @return [Gitlab::ObjectifiedHash] Information about created project.
     def create_project(name, options={})
+      if options.key?(:namespace_id) && options[:namespace_id].is_a?(String)
+        resp =
+            eval(get("/namespaces?search=#{options[:namespace_id]}"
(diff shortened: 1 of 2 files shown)


Sleeplessy:
A dirty code,but it works

lib/gitlab/client/projects.rb
```diff
@@ -87,9 +87,25 @@ def project_events(project, options={})
     # @option options [Integer] :user_id The user/owner id of a project.
     # @return [Gitlab::ObjectifiedHash] Information about created project.
     def create_project(name, options={})
+      if options.key?(:namespace_id) && options[:namespace_id].is_a?(String)
+        resp =
+            eval(get("/namespaces?search=#{options[:namespace_id]}"
```
NARKOZ: Don't use eval.

PR #355, opened 2018-02-06
Rename create_tag param

NaturalHokke:
https://docs.gitlab.com/ee/api/tags.html#create-a-new-tag

parameter name is `release_description`


NARKOZ:
Thank you.

Issue #361, opened 2018-02-15
Homebrew install


frob:
Add an install via homebrew.

I am not a ruby developer, I want to use this tool but I don't want to have to setup and maintain a ruby environment in order to use this tool. I don't use Ruby for anything else.

criztovyl:
Maybe check if [brew-gem](https://github.com/sportngin/brew-gem) for installing gems via homebrew suits your needs. :)

frob:
That worked, a proper homebrew formula would be nice, but this works. I have proposed adding this as an installation method in the readme.md #362

frob:
That worked for me, but in #362 @NARKOZ stated that a real homebrew formula would be better. 

So how do we do that?

NARKOZ:
I've submitted a PR to homebrew-core https://github.com/Homebrew/homebrew-core/pull/24421

PR #362, opened 2018-02-16
Update README.md to include homebrew installation method

frob:
Add homebrew gem as an installation option as recommended in #361


NARKOZ:
I'd prefer to see an official homebrew formula for it.

frob:
That would be best, but I currently don't have time to figure out how to do that. This I can manage.
Maybe (as a way of keeping perfect from being the enemy of the good) merge this but keep #361 open so movement can still be made to get a true homebrew formula.

PR #366, opened 2018-02-27
Improvement to ssh_keys, gemspec

otheus:
This patch provides support for maintaining the ssh_keys of other users. The existing implementation is not consistent in its use of user-id as a parameter; this is corrected when fetching keys. For these call implementations, pseudo-function-overloading is implemented with variable argument lists.

Also, a prior version introduced the use of yaml.safe_load, which is available only in psych +2.2, so that dependency is coded into the Gemspec.

It is so far still possible to build and deploy on Ruby 2.0 system (ie, RHEL). Let's keep it that way please.

### DIFF
--- a/gitlab.gemspec
+++ b/gitlab.gemspec
@@ -22,8 +22,8 @@ Gem::Specification.new do |gem|
 
   gem.add_runtime_dependency 'httparty'
   gem.add_runtime_dependency 'terminal-table'
+  gem.add_runtime_dependency 'psych',[">= 2.2.0"]
 
-  gem.add_development_dependency 'pry'
   gem.add_development_dependency 'rake'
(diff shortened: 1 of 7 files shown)


gitlab.gemspec
```diff
@@ -22,8 +22,8 @@ Gem::Specification.new do |gem|
 
   gem.add_runtime_dependency 'httparty'
   gem.add_runtime_dependency 'terminal-table'
+  gem.add_runtime_dependency 'psych',[">= 2.2.0"]
 
-  gem.add_development_dependency 'pry'
```
sotayamashita: I would appreciate if you could tell me a reason why did you delete "pry" library.

NARKOZ:
Closing since we support only Ruby >= 2.3.0 now.

Issue #356, opened 2018-02-06
Add the `events` API


johnmcconnell:

johnmcconnell:
From: lib/gitlab/client/issues.rb


```ruby
    def events(group=nil, options={})
      get("/groups/#{group}/issues", query: options)
    end

    def project_events(pid, options={})
      get("/projects/#{pid}/events", query: options)
    end

    def user_events(uid, options={})
      get("/users/#{uid}/events", query: options)
    end
```

abrophy:
Gave it a bash

PR #372, opened 2018-03-23
Add events API to the client

abrophy:


abrophy:
fixes #356

lib/gitlab/client/events.rb
```diff
@@ -0,0 +1,58 @@
...
+    # Gets a list of authenticated user's events
+    #
+    # @example
+    #   Gitlab.events()
+    #   Gitlab.events({ action: 'created', target_type: 'issue' })
+    #
+    # @param  [Hash] options A customizable set of options.
+    # @option options [String] :action Only events of specific action type
+    # @option options [String] :target_type Only events of specific target type
+    # @option options [Date] :before Only events created before date
```
johnmcconnell: Only thing I think of is are these dates going to be formatted correctly in the query params when in production?

I'll check to see if dates are used elsewhere

johnmcconnell: Example: https://github.com/NARKOZ/gitlab/blob/80960fab45dbf6c9c2b902bc930ab4f8d4d7ae4e/lib/gitlab/client/snippets.rb#L58

asedge:
Thanks!

PR #373, opened 2018-03-23
create_user explicitly requires username

### LINKED ISSUE #346: #create_user endpoint requires username
Gem version: 4.3.0

The Function `Gitlab.create_user` allows for a sparse call, with only email and password, but the server errors out since a username is required. Name is also required.
Line in question https://github.com/NARKOZ/gitlab/blob/v4.3.0/lib/gitlab/client/users.rb#L38
Docs https://docs.gitlab.com/ee/api/users.html#user-creation

HTTParty Error
```
=> #<HTTParty::Response:0x7ffe44ae0bb0 parsed_response=#<Gitlab::ObjectifiedHash:70364999105340 {hash: {"error"=>"username is missing"}}, @response=#<Net::HTTPBadRequest 400 Bad Request readbody==true>, @headers={"server"=>["nginx"], "date"=>["Fri, 12 Jan 2018 18:07:14 GMT"], "content-type"=>["application/json"], "content-length"=>["31"], "connection"=>["close"], "cache-control"=>["no-ccache"], "vary"=>["Origin"], "x-content-type-options"=>["nosniff"], "x-frame-options"=>["SAMEORIGIN"], "x-request-id"=>["0a3669bb-a4e6-4d37-aba0-2eaab1584328"], "x-runtime"=>["0.018132"]}>
```

abrophy:
fixes #346

### DIFF
--- a/lib/gitlab/client/users.rb
+++ b/lib/gitlab/client/users.rb
@@ -35,7 +35,7 @@ class Gitlab::Client
     # @example
     #   Gitlab.create_user('joe@foo.org', 'secret', 'joe', { name: 'Joe Smith' })
     #   or
-    #   Gitlab.create_user('joe@foo.org', 'secret')
+    #   Gitlab.create_user('joe@foo.org', 'secret', 'joe')
     #
     # @param  [String] email The email of a user.
     # @param  [String] password The password of a user.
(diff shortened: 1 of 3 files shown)


lib/gitlab/client/users.rb
```diff
@@ -50,9 +50,9 @@ def user(id=nil)
     def create_user(*args)
       options = Hash === args.last ? args.pop : {}
       body = if args[2]
-               { email: args[0], password: args[1], username: args[2] }
+               { email: args[0], password: args[1], username: args[2], name: args[0] }
              else
-               { email: args[0], password: args[1], name: args[0] }
+               raise ArgumentError.new("Missing required parameters")
              end
```
NARKOZ: Can we do `body = { email: args[0], password: args[1], username: args[2], name: args[0] }` instead? 🤔

NARKOZ:
ping @abrophy

Issue #310, opened 2017-08-08
snippets require visibility option


logicminds:
https://docs.gitlab.com/ee/api/snippets.html

The current docs in this gem for the snippets api do not show visibility as being required.

logicminds:
https://github.com/NARKOZ/gitlab/blob/master/lib/gitlab/client/snippets.rb#L31

NARKOZ:
Pull request is welcome

timofonic:
@NARKOZ Great, but would you consider interesting to provide some guidelines to possible contributor(s) about this? :)

Rubemlrm:
i thinks this issue can be closed already. what do you think ?

PR #380, opened 2018-05-01
Added visibility option into snippets

Rubemlrm:
Added visibility option into snippets documentation and specs. This was based on the issue #310


NARKOZ:
Thank you.

PR #383, opened 2018-05-22
URL encode Group ID

rspeicher:
The group ID can be a String, similar to a project ID, and so needs
to be encoded.


NARKOZ:
Thank you.

PR #390, opened 2018-06-14
Removed unnecessary files from Gemspec to reduce file size of distributed gem

zealot128:
Background: I want to use gitlab gem to run in a CLI and vendor all the dependencies. Gitlab was the largest with over 1MB. Granted, compared to node_modules the overhead is tiny, but nevertheless there are some needless configs that are shipped with the Gem but never needed for a usage.

Removed /specs + Docker configs + Dev configs

Size of the unzipped gem dir:
Before: 1,4MB 
After: 380KB (-72%)

Instead of reject { .. regex } one could use grep_v, which is available from Ruby 2.3. I was not sure, if you are supporting Ruby 2.2 and lower though.

### DIFF
--- a/gitlab.gemspec
+++ b/gitlab.gemspec
@@ -11,7 +11,9 @@ Gem::Specification.new do |gem|
   gem.summary       = 'A Ruby wrapper and CLI for the GitLab API'
   gem.homepage      = 'https://github.com/narkoz/gitlab'
 
-  gem.files         = `git ls-files`.split($/)
+  gem.files         = `git ls-files`.split($/).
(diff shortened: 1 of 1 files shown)


NARKOZ:
LGTM 👍 

>Instead of reject { .. regex } one could use grep_v, which is available from Ruby 2.3. I was not sure, if you are supporting Ruby 2.2 and lower though.

No, we don't support Ruby 2.2 anymore as it reached EOL. https://www.ruby-lang.org/en/downloads/branches/

NARKOZ:
Thank you.

PR #395, opened 2018-07-03
Added .unshare_project_with_group to projects

### LINKED ISSUE #394: Project unsharing is missing
Adding a sharing is possible but it seems the removal feature is missing: https://docs.gitlab.com/ee/api/projects.html#delete-a-shared-project-link-within-a-group

akkee:
Closes #394


akkee:
How about this @asedge ?

PR #399, opened 2018-08-03
Fixed options for projects api v4 and updated docs

### LINKED ISSUE #397: projects/:scope is deprecated on API v4 and scopes are now set on query parameters
Hi there 👋 
I was making calls to get a list of projects owned by the authenticated user, but I was getting a 404 error following the [example from the gem docs](https://www.rubydoc.info/gems/gitlab/Gitlab%2FClient%2FProjects:projects):
```ruby
client = Gitlab.client(endpoint: "https://gitlab.com/api/v4", private_token: access_token)
projects = lab.projects({scope: "owned"})
```

I looked into the [Gitlab API v4 docs](https://docs.gitlab.com/ce/api/projects.html#list-user-projects) and saw that `projects/:scope` was replaced by `owned`, `membership` and `starred` query parameters.

Indeed, the following query worked for me:
```ruby
projects = lab.projects({owned: true})
```

So, I think the IF clause that checks if there is a scope property can be removed from the `projects` method:
https://github.com/NARKOZ/gitlab/blob/b543f256c54be60a228346ba9f9932b62ffb6eba/lib/gitlab/client/projects.rb#L15-L21

And the docs can be updated to mention to use the new query parameters instead of the `scope`
https://github.com/NARKOZ/gitlab/blob/b543f256c54be60a228346ba9f9932b62ffb6eba/lib/gitlab/client/projects.rb#L13

If you agree with that I can submit a PR :)

mwebler:
Fixes #397

### DIFF
--- a/lib/gitlab/client/projects.rb
+++ b/lib/gitlab/client/projects.rb
@@ -10,14 +10,11 @@ class Gitlab::Client
     # @param  [Hash] options A customizable set of options.
     # @option options [Integer] :page The page number.
     # @option options [Integer] :per_page The number of results per page.
-    # @option options [String] :scope Scope of projects. 'owned' for list of projects owned by the authenticated user, 'all' to get all projects (admin only)
+    # (Any provided options will be passed to Gitlab. See {https://docs.gitlab.com/ce/api/projects.html#list-all-projects Gitlab docs} for all valid options)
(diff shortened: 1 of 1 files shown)


NARKOZ:
Thank you.

PR #401, opened 2018-08-09
Allow to pass variables to create_pipeline

lksv:


lib/gitlab/client/pipelines.rb
```diff
@@ -38,9 +38,21 @@ def pipeline(project, id)
...
     # @return [Gitlab::ObjectifiedHash] The pipelines changes.
-    def create_pipeline(project, ref)
-      post("/projects/#{url_encode project}/pipeline?ref=#{ref}")
+    def create_pipeline(project, ref, variables = {})
+      normalized_variables = variables.each_with_object([]) do |(key, value), res|
+        res.push(
+          'key' => key,
+          'value' => value
+        )
+      end
```
NARKOZ: Let variables be a string. We don't need this conversion logic in API wrapper.

NARKOZ:
Kind reminder @lksv

littldr:
Any progress here? I've stumbled over the same 'issue' and would love to see this feature merged. :)
Thanks for the good work!

lib/gitlab/client/pipelines.rb
```diff
@@ -38,9 +38,21 @@ def pipeline(project, id)
...
+
+      post(
+        "/projects/#{url_encode project}/pipeline",
+        query: { ref: ref },
+        body: JSON.dump(normalized_variables)
```
littldr: I've played around with this change and noticed that it does not work as expected. The problem is, that the API is missing the name of the parameter (`variables=...` see https://docs.gitlab.com/ee/api/pipelines.html#create-a-new-pipeline). Furthermore this gem sets the `Content-Type` header to `x-www-form-urlencoded` so the server does not understand the json structure which is sent in the body of the request.

The following code fixes these issues:
```ruby
def create_pipeline(project, ref, variables = {})
  # This is necessary, cause the API expects an array with objects (with `key` and `value` keys) <_< 
  normalized_variables = variables.each_with_object([]) do |(key, value), res|
    res.push(key: key, value: value)
  end

  post(
    "/projects/#{url_encode project}/pipeline",
    query: { ref: ref },
    body: { variables: normalized_variables },
  )
end
```

@NARKOZ do you prefer that this MR gets updated, or should I open another MR with this fix?

NARKOZ: @littldr please open a new one.

NARKOZ:
91afbd71df5fb8363bdd35e9ae620fc9a7cb3fcb

Issue #402, opened 2018-08-15
Missing API calls to manage project badges


fhaefemeier:
Since GitLab 10.6 a new feature is provided to manage [project badges](https://docs.gitlab.com/ee/api/project_badges.html). It can be done via GUI and API. This feature can be used to create or modify project badges during a CI build dynamically via API calls. Currently this API calls are not supported in NARKOZ/gitlab, but would be helpful.

fhaefemeier:
I am working on it and will prepare a pull request soon.

PR #404, opened 2018-08-16
Add API calls to manage group/project badges

fhaefemeier:
Enhancement #402 implemented. It add API functions to get, edit, add and remove badges of groups and projects. Unit tests implemented.


spec/gitlab/client/projects_spec.rb
```diff
@@ -66,8 +142,6 @@
     end
 
     it 'returns information about a created project' do
-      expect(@project.name).to eq('Gitlab')
-      expect(@project.owner.name).to eq('John Smith')
```
NARKOZ: Why are these deleted?

fhaefemeier: Don't knbow, it was a mistake. I add it back.

fhaefemeier:
I checked the lint job messages. I have trouble to fix the message:
`spec/gitlab/client/projects_spec.rb:20:86: C: Style/FormatStringToken: Prefer annotated tokens (like %<foo>s) over template tokens (like %{foo}).
      expect(@badges.first.link_url).to eq('http://example.com/ci_status.svg?project=%{project_path}&ref=%{default_branch}')
`. Do you know how it can be changed in a way to satisfy rubocop? I am not so familar with it. Is it possible to quote the string in some way, so it will no longer recognized as tokens?

NARKOZ:
ping @fhaefemeier

fhaefemeier:
> ping @fhaefemeier

Sorry, I hadn't time to fix the issues yet. I will try to do it in the next days.

PR #415, opened 2018-10-08
Add support for Version API

krames:
This PR adds support for the Version api that was added in GitLab 8.13.

https://docs.gitlab.com/ce/api/version.html

### DIFF
--- a/lib/gitlab/client/versions.rb
+++ b/lib/gitlab/client/versions.rb
@@ -1,0 +1,18 @@
+# frozen_string_literal: true
+
+class Gitlab::Client
+  # Defines methods related to version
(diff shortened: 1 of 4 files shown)


lib/gitlab/client/versions.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+
+class Gitlab::Client
+  # Defines methods related to version
+  # @see https://docs.gitlab.com/ce/api/version.html
+  module Versions
```
krames: I tried to follow the existing pattern for API calls. Please let me know if there is a more appropriate location for this method call.

NARKOZ:
Thank you.

PR #417, opened 2018-10-10
Added support for timeout in create_merge_request call

neeraj-joshi01:


NARKOZ:
Thank you for your pull request. I think this should be general and defined via httparty config 
https://github.com/NARKOZ/gitlab/blob/8e9e3212214249f02a4c59a728f929c8d6ff9639/lib/gitlab/configuration.rb#L47

NARKOZ:
If this is already possible with httparty options then a minor change to readme describing that option should be sufficient.

PR #427, opened 2018-10-31
Add 'Participants on issues' endpoint

brettmortensen:
Adds missing 'Participants on issues' endpoint and spec.

Documentation reference: https://docs.gitlab.com/ce/api/issues.html#participants-on-issues


NARKOZ:
Thank you ❤️

PR #428, opened 2018-11-02
Allow passing variables when creating a pipeline

littldr:
@NARKOZ as mentioned in #401, here is a new MR with the updated code. :)

Unfortunately the Gitlab API does not allow an empty `variables` parameter, so it is necessary to add this `if variables.any?` check... Maybe you know a better way how to ensure that the `variables` parameter is only set when at least one variable is passed?

### DIFF
--- a/lib/gitlab/client/pipelines.rb
+++ b/lib/gitlab/client/pipelines.rb
@@ -38,9 +38,20 @@ class Gitlab::Client
     #
     # @param  [Integer, String] project The ID or name of a project.
     # @param  [String] ref Reference to commit.
+    # @param  [Hash] variables Variables passed to pipelines
     # @return [Gitlab::ObjectifiedHash] The pipelines changes.
-    def create_pipeline(project, ref)
-      post("/projects/#{url_encode project}/pipeline?ref=#{ref}")
+    def create_pipeline(project, ref, variables = {})
+      body = {}
+
+      # This mapping is necessary, cause the API expects an array with objects (with `key` and `value` keys)
+      # See: https://docs.gitlab.com/ee/api/pipelines.html#create-a-new-pipeline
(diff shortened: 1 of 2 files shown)


lib/gitlab/client/pipelines.rb
```diff
@@ -38,9 +38,25 @@ def pipeline(project, id)
...
+      body = {}
+
+      if variables.any?
+        # This mapping is necessary, cause the API expects an array with objects (with `key` and `value` keys)
+        # See: https://docs.gitlab.com/ee/api/pipelines.html#create-a-new-pipeline
+        normalized_variables = variables.map do |(key, value)|
+          { key: key, value: value }
+        end
+        body[:variables] = normalized_variables
+      end
```
NARKOZ: How about this?

```rb
body[:variables] = variables.each_with_object([]) do |(key, value), arr|
  arr << { key: key, value: value }
end
```

littldr: It would work this way:
```ruby
body[:variables] = variables.map do |(key, value)|
  { key: key, value: value }
end if variables.any?
```

But I personally do not like the trailing condition in combination with multi-line statements...

littldr: The whole statement in one-line looks reasonable:
```ruby
body[:variables] = variables.map { |(key, value)| { key: key, value: value } } if variables.any?
```

What do you think?

NARKOZ: > What do you think?

lgtm

littldr: I've updated the PR! See 6a3f6e6

NARKOZ:
Thank you ❤️

PR #430, opened 2018-11-06
Fixes a bug in handling "Spam detected" messages

halfbyte:
This fixes `error.rb` so that it handles error responses created by Gitlab.com's spam protection
correctly. The error message is nested in a slightly off way (I believe
this is a bug), causing the error message formatting code to blow up as
it tries to join a string.

I admit doing `[val].flatten` feels a bit hackish, but the alternative (nested ternaries) is also not pretty. :)


NARKOZ:
Thank you.

PR #451, opened 2018-12-25
Protected tags API

### LINKED ISSUE #435: Protected Tags
Looks like this was added relatively recently: https://docs.gitlab.com/ce/api/protected_tags.html

The tags class has nothing for protected tags (maybe it should be a separate class?): https://www.rubydoc.info/gems/gitlab/Gitlab/Client/Tags

akkee:
**Protected tags API**:

```rb
  Gitlab.protected_tags(project)
  Gitlab.protected_tag(project, name)
  Gitlab.protect_repository_tag(project, name)
  Gitlab.unprotect_repository_tag(project, name)
```

closes #435

### DIFF
--- a/spec/gitlab/client/protected_tags_spec.rb
+++ b/spec/gitlab/client/protected_tags_spec.rb
@@ -1,0 +1,62 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+describe Gitlab::Client do
+  describe '.protected_tags' do
+    before do
+      stub_get('/projects/1/protected_tags', 'protected_tags')
+      @protected_tags = Gitlab.protected_tags(1)
+    end
+
+    it 'gets the correct resource' do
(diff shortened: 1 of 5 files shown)


spec/gitlab/client/protected_tags_spec.rb
```diff
@@ -0,0 +1,62 @@
...
+
+    it 'returns correct information about the protected repository tag' do
+      expect(@protected_tag.name).to eq 'release-1-0'
+    end
+  end
+
+  describe '.unprotect_repository_tag' do
+    before do
+      stub_delete('/projects/1/protected_tags/release-1-0', 'empty')
+      @unprotected_tag = Gitlab.unprotect_repository_tag(1, 'release-1-0')
```
NARKOZ: This variable doesn't seem to be used elsewhere. It could be removed.

Issue #458, opened 2019-01-25
Http error 429


MarkGibbons:
When rate limiting is turned on in gitlab and a 429 is returned the body is returned as text, not json.
An error "Gitlab::Error::Parsing: The response is not a valid JSON" is raised instead of a TooManyRequests Retry Later error.

I got this stack trace.
Gitlab::Error::Parsing: chef_cookbook[jenkins] (restaurant::cookbooks line 52) had an error: Gitlab::Error::Parsing: gitlab_hook[jenkins org webhook] (/var/chef/cache/cookbooks/restaurant/resources/chef_cookbook.rb line 136) had an error: Gitlab::Error::Parsing: The response is not a valid JSON
/opt/chef/embedded/lib/ruby/gems/2.4.0/gems/gitlab-4.6.1/lib/gitlab/request.rb:40:in `rescue in decode'
/opt/chef/embedded/lib/ruby/gems/2.4.0/gems/gitlab-4.6.1/lib/gitlab/request.rb:36:in `decode'
/opt/chef/embedded/lib/ruby/gems/2.4.0/gems/gitlab-4.6.1/lib/gitlab/request.rb:18:in `parse'
/opt/chef/embedded/lib/ruby/gems/2.4.0/gems/gitlab-4.6.1/lib/gitlab/request.rb:12:in `block in <class:Request>'
/opt/chef/embedded/lib/ruby/gems/2.4.0/gems/httparty-0.16.2/lib/httparty/request.rb:391:in `parse_response'
/opt/chef/embedded/lib/ruby/gems/2.4.0/gems/httparty-0.16.2/lib/httparty/request.rb:359:in `block in handle_response'
/opt/chef/embedded/lib/ruby/gems/2.4.0/gems/httparty-0.16.2/lib/httparty/response.rb:25:in `parsed_response'
/opt/chef/embedded/lib/ruby/gems/2.4.0/gems/gitlab-4.6.1/lib/gitlab/error.rb:43:in `build_error_message'
/opt/chef/embedded/lib/ruby/gems/2.4.0/gems/gitlab-4.6.1/lib/gitlab/error.rb:20:in `initialize'
/opt/chef/embedded/lib/ruby/gems/2.4.0/gems/gitlab-4.6.1/lib/gitlab/request.rb:68:in `exception'
/opt/chef/embedded/lib/ruby/gems/2.4.0/gems/gitlab-4.6.1/lib/gitlab/request.rb:68:in `raise'
[...]

PR #462, opened 2019-01-28
Add support to handle 429 responses from a rate limited gitlab instance

MarkGibbons:
Gitlab may respond with a 429 error and a message of 'Retry later'.  The
response body in this case has content-type 'text/plain'. This change
adds support for 429 errors and processes the body of responses with
content formated as 'text/plain'.

Should fix issue #458.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -14,6 +14,9 @@ Metrics/BlockLength:
   Exclude:
     - 'spec/**/*'
 
+Metrics/CyclomaticComplexity:
+  Max: 15
+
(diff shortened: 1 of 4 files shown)


NARKOZ:
Thanks for the PR. Could you add tests for this change?

MarkGibbons:
I'll try.

MarkGibbons:
I added two tests to exercise the method I added.  I sorry they aren't very good, I don't know how to create a httparty::response for testing.

NARKOZ:
Thanks!

PR #343, opened 2017-12-15
Add group-issues API

dosuken123:

PR #465, opened 2019-01-31
Add group-issues API (redux)

SeanSith:
Following up on the feedback in #343 fixed the failing test.


lib/gitlab/client/groups.rb
```diff
@@ -191,5 +191,17 @@ def group_subgroups(id, options = {})
...
+
+    # Gets a list of issues of a group.
+    #
+    # @example
+    #   Gitlab.group_issues(5)
+    #
+    # @param  [Integer, String] group_id The ID or name of a group.
+    # @param  [Hash] options A customizable set of options.
+    # @return [Array<Gitlab::ObjectifiedHash>]
+    def group_issues(group = nil, options = {})
```
NARKOZ: First argument (group) is required therefore shouldn't have a default value.

NARKOZ:
Thank you!

PR #475, opened 2019-02-07
Add support for runner jobs status filter

jindrichskupa:


lib/gitlab/client/runners.rb
```diff
@@ -79,9 +79,11 @@ def delete_runner(id)
     #   Gitlab.runner_jobs(1)
     #
     # @param  [Integer] id The ID of a runner.
+    # @param  [String] status The job status filter
     # @return [Array<Gitlab::ObjectifiedHash>]
-    def runner_jobs(runner_id)
-      get("/runners/#{url_encode runner_id}/jobs")
+    def runner_jobs(runner_id, status=nil)
+      filter = status.nil? ? "" : "?status=#{status}"
+      get("/runners/#{url_encode runner_id}/jobs#{filter}")
```
NARKOZ: We should pass a hash of options as a query. See the line 60 for an example.

NARKOZ:
#492

PR #488, opened 2019-02-24
Added missing endpoints in Runners API

### LINKED ISSUE #474: Verify and Register Runner
https://docs.gitlab.com/ee/api/runners.html#verify-authentication-for-a-registered-runner

It seems this is missing from `runners.rb`: https://github.com/NARKOZ/gitlab/blob/43e33224d529f591247a9993ea9bd11e17c79d06/lib/gitlab/client/runners.rb

It also seems to be missing Register Runner (https://docs.gitlab.com/ee/api/runners.html#register-a-new-runner)

akkee:
```ruby
  Gitlab.register_runner(token, options = {})
  Gitlab.delete_registered_runner(token)
  Gitlab.verify_auth_registered_runner(token)
```

Closes #474


lib/gitlab/client/runners.rb
```diff
@@ -122,5 +122,50 @@ def project_enable_runner(project_id, id)
...
+      post('/runners', body: body)
+    end
+
+    # Deletes a registed Runner.
+    #
+    # @example
+    #   Gitlab.delete_registered_runner('9142c16ea169eaaea3d752313a434a6e')
+    #
+    # @param  [String] token Runner authentication token.
+    # @return [void] This API call returns an empty response body.
```
NARKOZ: `@return [nil]`

lib/gitlab/client/runners.rb
```diff
@@ -122,5 +122,50 @@ def project_enable_runner(project_id, id)
...
+      delete('/runners', body: body)
+    end
+
+    # Validates authentication credentials for a registered Runner.
+    #
+    # @example
+    #   Gitlab.verify_auth_registered_runner('9142c16ea169eaaea3d752313a434a6e')
+    #
+    # @param  [String] token Runner authentication token.
+    # @return [void] This API call returns an empty response body.
```
NARKOZ: `@return [nil]`

lib/gitlab/client/runners.rb
akkee: Done

lib/gitlab/client/runners.rb
akkee: Done

PR #490, opened 2019-03-01
Add the gitlab CI variable as a fallback for Gitlab.endpoint

alexives:
Gitlab CI Provides CI_API_V4_URLfor the V4 API as of 11.7. It'd be nice to just pick that up if using the gem in a gitlab ci job.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,7 +40,7 @@ Configuration example:
 
 ```ruby
 Gitlab.configure do |config|
-  config.endpoint       = 'https://example.net/api/v4' # API endpoint URL, default: ENV['GITLAB_API_ENDPOINT']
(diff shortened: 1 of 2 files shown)


NARKOZ:
Not sure about this change. GitLab CI allows you set env vars https://docs.gitlab.com/ee/ci/variables/#gitlab-ciyml-defined-variables so setting `GITLAB_API_ENDPOINT` variable shouldn't be a problem.

baurmatt:
I like the idea! :) That way you don't have have to specify the `GITLAB_API_ENDPOINT` variable in CI jobs. It just works out of the box.

baurmatt:
I definitely voting to merge this feature! :)

@NARKOZ do you mind taking another look at this PR? Thanks! :)

PR #492, opened 2019-03-05
Add options to runner_jobs

atiaxi:
Ran into an issue where I needed to pass options to `runner_jobs`, so I've added the ability to do so.  Similar to #475 in spirit but passes through the full `options` hash rather than just a `status` argument.  Also updated the tests to include passing in the option.


NARKOZ:
Thank you ❤️

PR #493, opened 2019-03-20
Edit approvals/approvers fix

Quantu:
Fix edit_project_approvers to use correct API path & update comments to properly reflect options as required, and to set to nil if you want one or both of them empty.


NARKOZ:
Thank you ❤️

PR #497, opened 2019-05-13
Added www in front of rubydoc.info

mbogh:
Accessing https://rubydoc.info/ gives an SSL error since their certificate only covers `www.rubydoc.info`.

Screenshot 2019-05-13 at 13 17 55

Screenshot 2019-05-13 at 13 22 09


NARKOZ:
Thank you ❤️

PR #498, opened 2019-05-15
Allow (un)archive of projects

baurmatt:
This PR implemented the needed method for (un)archiving a project.


NARKOZ:
Thank you ❤️

PR #499, opened 2019-05-23
Add a paginate_with_limit pagination function, to paginate but return…

craigmiskell-gitlab:
… only up to X results

Useful for situations like fetching the 200 most recent MRs from a large project.


craigmiskell-gitlab:
Tests added. Not sure what to do about the cognitive complexity, if anything.

NARKOZ:
Thank you ❤️

Issue #494, opened 2019-04-25
Wrong path for MR endpoint


reshadf:
### Summary

When using the Ruby Wrapper for the GITLAB REST API, fetching the MR discussions fails. This is probably because the Request URI is not correct. ( see below )

### Steps to reproduce

within Danger ( danger.systems )

call the following API 
`client.merge_request_discussions(ci_source.repo_slug, ci_source.pull_request_id).auto_paginate`

### What is the current *bug* behavior?

(What actually happens)
The current output is generated because the call fails due to the URL not being correct in the Ruby Gems API Wrapper (v4.11.0) : 
```
Gitlab::Error::NotFound: Server responded with code 404, message: 404 API Version Not Found. Request URI: https://gitlab.intern.mydomain.com/api/v4http://gitlab.intern.mydomain.com/api/v4/projects/26/merge_requests/443/discussions?id=26¬eable_id=443&page=2&per_page=20
  /usr/share/rvm/gems/ruby-2.6.2/gems/gitlab-4.11.0/lib/gitlab/request.rb:67:in `validate'
  /usr/share/rvm/gems/ruby-2.6.2/gems/gitlab-4.11.0/lib/gitlab/request.rb:46:in `block (2 levels) in <class:Request>'
  /usr/share/rvm/gems/ruby-2.6.2/gems/gitlab-4.11.0/lib/gitlab/paginated_response.rb:92:in `next_page'
  /usr/share/rvm/gems/ruby-2.6.2/gems/gitlab-4.11.0/lib/gitlab/paginated_response.rb:40:in `each_page'
  /usr/share/rvm/gems/ruby-2.6.2/gems/gitlab-4.11.0/lib/gitlab/paginated_response.rb:47:in `auto_paginate'
  /usr/share/rvm/gems/ruby-2.6.2/gems/danger-6.0.6/lib/danger/request_sources/gitlab.rb:78:in `mr_comments'
  /usr/share/rvm/gems/ruby-2.6.2/gems/danger-6.0.6/lib/danger/request_sources/gitlab.rb:172:in `update_pull_request_with_inline_comments!'
  /usr/share/rvm/gems/ruby-2.6.2/gems/danger-6.0.6/lib/danger/request_sources/gitlab.rb:165:in `update_pull_request!'
  /usr/share/rvm/gems/ruby-2.6.2/gems/danger-6.0.6/lib/danger/danger_core/dangerfile.rb:247:in `post_results'
[...]

PR #500, opened 2019-05-24
paginated_response regex tweak

Quantu:
This _should_ be a working fix for issue #494 (Wrong path for MR endpoint).

### DIFF
--- a/lib/gitlab/paginated_response.rb
+++ b/lib/gitlab/paginated_response.rb
@@ -56,6 +56,10 @@ module Gitlab
       response
     end
 
+    def parse_path(link)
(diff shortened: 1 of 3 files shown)


lib/gitlab/paginated_response.rb
```diff
@@ -56,6 +56,10 @@ def auto_paginate
       response
     end
 
+    def parse_path(link)
+      link[%r{\/(?!.*\.).*}].sub(/#{@client.endpoint[%r{\/(?!.*\.).*}]}/, '')
```
NARKOZ: Could you add some comments explaining the regex?

PR #501, opened 2019-05-25
Add format option to repo_archive

excid3:
This PR adds support for choosing which archive format when downloading the archive for a repository. 

This follows the docs here: https://docs.gitlab.com/ee/api/repositories.html#get-file-archive


NARKOZ:
Thank you ❤️

PR #505, opened 2019-06-24
Raise exception on HTTP status 406

theflow:
I've ran into this when [accepting a MR](https://docs.gitlab.com/ce/api/merge_requests.html#accept-mr), which can return a 406 status code.


NARKOZ:
Thank you ❤️

PR #508, opened 2019-07-25
Updating Repositories code documentation

FranciscoAmado:
Hi there 👋 

While using the `tree`  method of Repositories I noticed one small discrepancy with the actual API.

Although it's not documented in code, there's one other option of passing the `per_page` value to configure how many results you want per page.

See https://docs.gitlab.com/ee/api/repositories.html#list-repository-tree

I've used this with the actual version of this wrapper and it's working 👍 

It's just a matter of updating the documentation.

Thanks for this wrapper! It's super useful 😄


asedge:
Thank you!

PR #512, opened 2019-08-02
search_in_project: add ref filter if provided - backward compatible w…

bmedici:
Gitlab API offers a "ref" parameter we can use to specify a ref when searching a repository.
Added this to  search_in_project, with a backward compatible fourth parameter

https://gitlab.com/gitlab-org/gitlab-ce/issues/46469
https://gitlab.com/gitlab-org/gitlab-ce/commit/a73ee1cc3cc78f1f25034be9394e980ff5209ecc


NARKOZ:
Thank you ❤️

PR #513, opened 2019-08-08
Allow custom header arguments via `httparty` option

rspeicher:
The `Gitlab::Client` initializer allows you to pass an `httparty` option
to provide additional options to the internal HTTParty client. This is great!
However, previously the `Gitlab::Request` class would overwrite the
`headers` value in order to set its authorization token.

Now, we merge the authorization token with any user-defined HTTParty
header arguments.


README.md
```diff
@@ -72,8 +72,14 @@ ENV['GITLAB_API_HTTPARTY_OPTIONS'] = '{read_timeout: 60}'
...
 
-# initialize a new client
-g = Gitlab.client(endpoint: 'https://api.example.com', private_token: 'qEsq1pt6HJPaNciie3MG')
+# initialize a new client with custom headers
+g = Gitlab.client(
+  endpoint: 'https://example.com/api/v4',
+  private_token: 'qEsq1pt6HJPaNciie3MG',
+  httparty: {
```
rspeicher: I didn't see the `httparty` option documented in the README at all, and I was pleasantly surprised to find it, so I thought it made sense to expose this new feature here.

lib/gitlab/request.rb
```diff
@@ -62,28 +67,25 @@ def validate(response)
     # Sets a base_uri and default_params for requests.
     # @raise [Error::MissingCredentials] if endpoint not set.
     def request_defaults(sudo = nil)
-      self.class.default_params sudo: sudo
       raise Error::MissingCredentials, 'Please set an endpoint to API' unless @endpoint
 
+      self.class.default_params sudo: sudo
```
rspeicher: I moved this below the `raise` while I was debugging some cross-test pollution; see https://github.com/NARKOZ/gitlab/pull/513/files#diff-a36c72b98dfda3970f82ee56454344e3R12-R13.

I'm not sure why `self.class.default_params(sudo: sudo) unless sudo.nil?` didn't work, but oh well. 🤷‍♂

spec/gitlab/request_spec.rb
```diff
@@ -7,7 +7,12 @@
   it { is_expected.to respond_to :post }
   it { is_expected.to respond_to :put }
   it { is_expected.to respond_to :delete }
+
   before do
+    # Prevent tests modifying the `default_params` value from causing cross-test
+    # pollution
+    described_class.default_params.delete(:sudo)
```
rspeicher: I was seeing order-dependent spec failures in this file depending on when the test on line 56 ran. This resolves that.

rspeicher:
@NARKOZ Could you please take a look at this one? 💜

We're doing some work with `canary.gitlab.com`, and the "correct" way of accessing the installation is via a custom `gitlab_canary` cookie, which didn't work with the current gitlab gem. See https://gitlab.com/gitlab-com/chatops/merge_requests/70#note_200964593 for more information.

NARKOZ:
Thank you ❤️

PR #521, opened 2019-10-02
Fix paginated_response when endpoint url diff from link header

azhuravlev:
Sometimes gitlab is requested via ip, but in response returned url.
For this case 'sub' doesn't work as expected and for this issue
we need to parse returned url and merge it with original gitlab endpont.

Fixes issue [494](https://github.com/NARKOZ/gitlab/issues/494)


NARKOZ:
Thanks for the PR. Could you add a test case?

PR #526, opened 2019-11-14
Options hash documentation for tree and commits methods is outdated

loperaja:
Gitlab API v4 uses the get parameter `ref` instead of `ref_name` when listing commits and files. Current documentation does not reflect that change. Its confusing because when `ref_name` is used, API will just default to master. We tested it in our server (master contains 9 files and version 1.0.0 has 7):

```
irb(main):061:0> g.tree(5419,ref_name: '1.0.0').count
=> 9
irb(main):062:0> g.tree(5419,ref: '1.0.0').count
=> 7
```

### DIFF
--- a/lib/gitlab/client/commits.rb
+++ b/lib/gitlab/client/commits.rb
@@ -8,11 +8,11 @@ class Gitlab::Client
     #
     # @example
     #   Gitlab.commits('viking')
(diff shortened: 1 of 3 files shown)


NARKOZ:
Thank you ❤️

Issue #527, opened 2019-11-20
Feature request: Add the ability to provide optional parameters on create and update BuildVariables


mrosePTC:
As the title states, the `Gitlab::Client::BuildVariables` module lacks the ability to provide optional parameters to `create_group_variable`, `create_variable`, `update_group_variable` and `update_variable`. The additional parameters described in the group variable documentation [here](https://docs.gitlab.com/ce/api/group_level_variables.html) and the project variable documentation [here](https://docs.gitlab.com/ee/api/project_level_variables.html) are not accessible with the current state on master.

I propose a change to each of the above methods that is something similar to this:

``` ruby
def create_group_variable(group, key, value, **opts)
    post("/projects/#{url_encode project}/variables", body: opts.merge({ key: key, value: value }))
end
```

The addition of this functionality would allow for the inclusion of all current parameters, and automatically support new parameters as they are introduced. Users would then be able to specify parameters like `protected` or `masked` without having to go to the gitlab UI.

If this approach is acceptable, I would be willing to provide a submission making these changes.

NARKOZ:
PR is welcome

mrosePTC:
This can be closed from my perspective

PR #528, opened 2019-11-22
Allow for optional parameters to be passed into BuildVariables

mrosePTC:
Also update URIs for gitlab docs concerning group and project
build variables.

This addresses https://github.com/NARKOZ/gitlab/issues/527


NARKOZ:
Thank you ❤️

PR #530, opened 2019-11-25
Forward missing messages on Error to the parsed response

rspeicher:
An error response may contain useful additional context, and this allows
users to access that context on the exception class.

For example, the cherry-pick API now includes an `error_code` field to
indicate why the cherry-pick failed, which can now be accessed via
`Error#error_code`.


NARKOZ:
👍

PR #539, opened 2019-12-26
Fix adding time spent with negative values

### LINKED ISSUE #538: Adding spent time with negative values cause timelogs error
Gitlab API supports providing negative values to add spent time endpoint which allows to remove all or part of the already tracked time. Since `add_time_spent_on_issue` method uses `url_encode` on `duration` before sending it to Gitlab, it results in error. `-2m` is sent as `%2d2m` and it returns `'timelogs' is invalid` error from GitLab.

Is it really necessary to use `url_encode` for duration?

lowski:
It fixes #538


NARKOZ:
Thank you ❤️

PR #540, opened 2020-01-10
Adds support to add/delete approval rules and get the approval settings

AngelVlc:
I've added these methods:

Related to approval rules:
- delete_approval_rule
- create_approval_rule

And related to approval settings:
- approval_settings_for_project

Thanks!

### DIFF
--- a/lib/gitlab/client/approval_rules.rb
+++ b/lib/gitlab/client/approval_rules.rb
@@ -1,4 +1,5 @@
 # frozen_string_literal: true
+
 require 'json'
 
 class Gitlab::Client
@@ -33,8 +34,7 @@ class Gitlab::Client
     # @return [Array<Gitlab::ObjectifiedHash>]
     def create_approval_rule(project, rule_name, options = {})
       body = { name: rule_name }.merge(options)
-      settings = post("/projects/#{url_encode project}/approval_settings/rules", body: body)
+      post("/projects/#{url_encode project}/approval_settings/rules", body: body)
     end
-
   end
 end
--- a/lib/gitlab/client/approval_rules.rb
+++ b/lib/gitlab/client/approval_rules.rb
@@ -1,0 +1,40 @@
+# frozen_string_literal: true
+require 'json'
+
+class Gitlab::Client
+  # Defines methods related to project approval rules.
+  # Does not exist in API documentation
+  module ApprovalRules
+    # Deletes the approval role from the project approval settings. Approval rules come within the approval settings.
+    #
+    # @example
+    #   Gitlab.delete_approval_rule(1, 133)
+    #   Gitlab.delete_approval_rule("project", 133)
+    #
+    # @param  [Integer, String] id The ID or name of a project.
+    # @param  [Integer, String] id The ID of the rule to be deleted.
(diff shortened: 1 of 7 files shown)


lib/gitlab/client/approval_rules.rb
```diff
@@ -0,0 +1,40 @@
+# frozen_string_literal: true
+
+require 'json'
```
NARKOZ: why is it required?

lib/gitlab/client/approval_rules.rb
```diff
@@ -0,0 +1,40 @@
...
+    end
+
+    # Crates an approval rule with the name and settings provided
+    #
+    # @example
+    #   Gitlab.create_approval_rule(1, "Default",  { :approvals_required => 2, :users =>[], :groups => [1, 2], :remove_hidden_groups => false })
+    #   Gitlab.delete_approval_rule("project", "Default", { :approvals_required => 2, :users =>[], :groups => [1, 2], :remove_hidden_groups => false })
+    #
+    # @param  [Integer, String] id The ID or name of a project.
+    # @param  [Integer, String] id The ID of the rule to be deleted.
```
NARKOZ: ✂️

lib/gitlab/client/approval_rules.rb
AngelVlc: It isn't required, I've removed.

lib/gitlab/client/approval_rules.rb
AngelVlc: Fixed.

lib/gitlab/client/approval_rules.rb
```diff
@@ -0,0 +1,38 @@
+# frozen_string_literal: true
+
+class Gitlab::Client
+  # Defines methods related to project approval rules.
+  # Does not exist in API documentation
```
NARKOZ: How did you write the code if there are no docs?

AngelVlc: I didn't write the code but my mates told me that they found them by inspecting the calls from the UI with Chrome 😅

NARKOZ:
I'm not sure that the client should support undocumented (unofficial) API.

NARKOZ:
Sorry, I'm closing this. This client wouldn't support unofficial API.

AngelVlc:
👍

johnbatty:
I was just looking at implementing...
https://docs.gitlab.com/12.8/ee/api/merge_request_approvals.html#get-project-level-rules
Which looks like (part of) the function this PR is trying to implement.
Would it be best to tweak this PR to use this API?

Fyi, I want support for this "project-level-rules" API as I have been trying to use the existing client API [project_merge_request_approvals](https://www.rubydoc.info/gems/gitlab/Gitlab/Client/MergeRequestApprovals#project_merge_request_approvals-instance_method) that calls "api/v4/projects/<id>/approvals) that calls "api/v4/projects/<id>/approvals", and found that the server has a bug that returns incorrect information (I've got an open GitLab issue for this).  Whereas the "get-project-level-rules" works, but this is not currently exposed via the Ruby client library.

PR #541, opened 2020-01-14
Add a method to delete job artifacts

javanthropus:
This adds a method for deleting job artifacts.

If you never set expiration for your artifacts and now need a mass delete, this function will help.


NARKOZ:
Thank you ❤️

PR #542, opened 2020-01-14
Avoid modifying request parameters in-place

rspeicher:
Consider the following script:

```ruby
require 'gitlab'

client = Gitlab::Client.new(
  endpoint: 'https://gitlab.com/api/v4',
  private_token: '[REDACTED]'
)

args = { per_page: 10 }
puts args.inspect

client.get('/projects', args)
puts args.inspect
```

Prior to this change, this would output:

    {:per_page=>10}
    {:per_page=>10, :headers=>{"PRIVATE-TOKEN"=>"[REDACTED]"}}

Which would be a problem if a user were logging the parameters they were
passing to a `PUT` endpoint inside a loop, for example. The first log
message would have their parameters, while subsequent logs would have
their parameters plus their private token.

This is unexpected behavior that could expose credentials.

Now we duplicate the user parameters before modifying them in-place.

### DIFF
--- a/lib/gitlab/request.rb
+++ b/lib/gitlab/request.rb
@@ -41,14 +41,16 @@ module Gitlab
 
     %w[get post put delete].each do |method|
       define_method method do |path, options = {}|
-        httparty_config(options)
+        params = options.dup
 
-        unless options[:unauthenticated]
-          options[:headers] ||= {}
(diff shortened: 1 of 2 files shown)


NARKOZ:
Thanks.

PR #546, opened 2020-02-02
add ruby 2.7 support

fe80:
Hello,

Since ruby 2.7 I've this warning:
```
gems/gitlab-4.13.1/lib/gitlab/client.rb:77: warning: URI.escape is obsolete
```

I suppose that now it's [encode_www_form_component](https://docs.ruby-lang.org/en/2.7.0/URI.html#method-c-encode_www_form_component)

### DIFF
--- a/lib/gitlab/client.rb
+++ b/lib/gitlab/client.rb
@@ -74,7 +74,7 @@ module Gitlab
     end
 
     def url_encode(url)
-      URI.encode(url.to_s, /\W/)
+      URI.encode_www_form_component(url.to_s)
     end
 
     private


NARKOZ:
> I suppose that now it's [encode_www_form_component](https://docs.ruby-lang.org/en/2.7.0/URI.html#method-c-encode_www_form_component)

AFAIK they're not interchangeable and behave differently. The build is failing because of that.

fe80:
For the query string ?

NARKOZ:
Support for Ruby 2.7 was added in d9ef580. A new version will be released soon. Thank you.

PR #555, opened 2020-03-23
Implement GET /user/activities

joeyates:
Adds `Gitlab.activities`, valid for admin auth tokens only: https://docs.gitlab.com/ee/api/users.html#get-user-activities-admin-only

### DIFF
--- a/lib/gitlab/client/users.rb
+++ b/lib/gitlab/client/users.rb
@@ -123,6 +123,20 @@ class Gitlab::Client
       post('/session', body: { email: email, password: password }, unauthenticated: true)
     end
 
+    # Gets a list of user activities (for admin access only).
+    #
+    # @example
+    #   Gitlab.activities
+    #
+    # @param  [Hash] options A customizable set of options.
+    # @option options [Integer] :page The page number.
+    # @option options [Integer] :per_page The number of results per page.
+    # @option options [DateTime] :from The start date for paginated results.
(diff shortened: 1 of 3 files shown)


joeyates:
Codeclimate is failing with `/usr/local/lib/ruby/site_ruby/2.4.0/rubygems/core_ext/kernel_require.rb:122:in `require': cannot load such file -- rubocop-performance`

lib/gitlab/client/users.rb
```diff
@@ -123,6 +123,20 @@ def session(email, password)
...
 
+    # Gets a list of user activities (for admin access only).
+    #
+    # @example
+    #   Gitlab.activities
+    #
+    # @param  [Hash] options A customizable set of options.
+    # @option options [Integer] :page The page number.
+    # @option options [Integer] :per_page The number of results per page.
+    # @option options [DateTime] :from The start date for paginated results.
```
NARKOZ:
```suggestion
    # @option options [String] :from The start date for paginated results.
```

>Date string in the format YEAR-MONTH-DAY. For example, 2016-03-11. Defaults to 6 months ago.

https://docs.gitlab.com/ce/api/users.html#get-user-activities-admin-only

joeyates: Fixed

NARKOZ:
Thank you 🖤

PR #557, opened 2020-04-13
Delegate Hash methods in ObjectifiedHash

rspeicher:
Assuming this class is meant to mimic a normal Ruby Hash if the sent
message doesn't exist as a data key, we need to also delegate methods
that the normal Hash responds to (e.g., `include?`).

This also corrects the method signature for `method_missing` to include
the remaining arguments and optional block.

Prior to this change, sending a message to ObjectifiedHash that wasn't one of
its keys, you'd get:

    ArgumentError:
      wrong number of arguments (given 2, expected 1)
    # ./lib/gitlab/objectified_hash.rb:27:in `method_missing'

### DIFF
--- a/lib/gitlab/objectified_hash.rb
+++ b/lib/gitlab/objectified_hash.rb
@@ -7,7 +7,7 @@ module Gitlab
     def initialize(hash)
       @hash = hash
       @data = hash.each_with_object({}) do |(key, value), data|
-        value = ObjectifiedHash.new(value) if value.is_a? Hash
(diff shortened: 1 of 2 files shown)


NARKOZ:
It wasn't intended to behave like a hash. That's why there is `to_hash` method.

renehernandez:
Chiming in. I am trying to delete ssh keys for the authenticated user and the `Gitlab.delete_ssh_key(id)` method fails with above mentioned error. Manually replacing the `method_missing` code with the proposed one in this PR fixes the issue for me

PR #561, opened 2020-05-11
Epics API: fetch and create epic notes + fetch epic issues

lienvdsteen:
Added two methods related to epics:

- #epic_notes: Fetch all the notes on an epic
- #create_epic_note: Add a note on an epic
- #epic_issues: Fetch all the issues for an epic


NARKOZ:
Thank you ❤️

PR #567, opened 2020-06-17
Add support for nested arrays on ObjectifiedHash

fabio-gallonetto:
At the moment ObjectifiedHash is missing out on hierarchies of objects where there are hashes contained within an array. Which is annoying as it happens in a lot of API calls.

This just adds an extra step to the initialiser of ObjectifiedHash so that arrays of hashes are handled properly


NARKOZ:
Thank you. Could you add a test for this change?

fabio-gallonetto:
Done!

NARKOZ:
Tests are failing 😢

fabio-gallonetto:
Something was wrong with GitHub today.. I closed the PR by mistake and can't open it again :(

NARKOZ:
Unfortunately no, the reopen button is disabled for me. Recreating a new PR shouldn't be a problem though.

PR #568, opened 2020-06-22
Simplify pagination using Enumerator::Lazy

javanthropus:
This preserves compatibility with the existing pagination methods while leveraging a lazy enumerator to simplify the implementation.  The other advantage is that using the new `lazy_paginate` method directly allows users to take advantage of all the additional methods available to lazy enumerators, such as `take` or `first`, without needing to code those methods into this library.

### DIFF
--- a/lib/gitlab/paginated_response.rb
+++ b/lib/gitlab/paginated_response.rb
@@ -42,37 +42,18 @@ module Gitlab
       end
     end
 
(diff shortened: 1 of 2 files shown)


javanthropus:
The remaining Rubocop failure is not in any code that I modified.

NARKOZ:
Great! Thank you.

PR #570, opened 2020-06-30
Add support for nested arrays on ObjectifiedHash (take 2)

fabio-gallonetto:
At the moment ObjectifiedHash is missing out on hierarchies of objects where there are hashes contained within an array. Which is annoying as it happens in a lot of API calls.

This just adds an extra step to the initialiser of ObjectifiedHash so that arrays of hashes are handled properly

### DIFF
--- a/lib/gitlab/objectified_hash.rb
+++ b/lib/gitlab/objectified_hash.rb
@@ -8,7 +8,7 @@ module Gitlab
       @hash = hash
       @data = hash.each_with_object({}) do |(key, value), data|
         value = self.class.new(value) if value.is_a? Hash
-        value = value.map { |v| self.class.new(v) } if value.is_a? Array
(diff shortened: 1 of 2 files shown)


fabio-gallonetto:
I got it working now: the problem was that turning on support for `hash.array[2].a` broke the legacy access mode `hash.array[2]['a']`.  So I've implemented support on the ObjectifiedHash for the subscript operator to support both the old and the new mode of accessing (to avoid a breaking change)

lib/gitlab/objectified_hash.rb
```diff
@@ -23,6 +24,10 @@ def inspect
       "#<#{self.class}:#{object_id} {hash: #{hash.inspect}}"
     end
 
+    def [](key)
+      @data[key]
```
NARKOZ, line 28:
```suggestion
      data[key]
```

NARKOZ:
Thank you!

PR #577, opened 2020-07-17
Support error_code returned by specific API endpoints

rspeicher:
An error response may contain useful additional context, and this allows
users to access that context on the exception class.

For example, the cherry-pick API now includes an error_code field to
indicate why the cherry-pick failed, which can now be accessed via
`ResponseError#error_code`.

This is a second attempt at https://github.com/NARKOZ/gitlab/pull/530, without abusing `method_missing`.

### DIFF
--- a/lib/gitlab/error.rb
+++ b/lib/gitlab/error.rb
@@ -34,6 +34,17 @@ module Gitlab
         @response.parsed_response.message
       end
 
+      # Additional error context returned by some API endpoints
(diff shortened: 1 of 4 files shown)


rspeicher:
@NARKOZ At your convenience, could we please get a new gem release to include this and https://github.com/NARKOZ/gitlab/pull/579 (among others)? 💜

Issue #581, opened 2020-09-09
Get group runners


krzaczek:
The library is missing a method to list all group runners

https://docs.gitlab.com/ee/api/runners.html#list-groups-runners

PR #583, opened 2020-09-12
Fixed instances where HTTP methods don't match the type of params

akashdotsrivastava:
While working on #581, I noticed a mismatch where a `query` was passed instead of a `body` in a `put` request. I fixed it in https://github.com/NARKOZ/gitlab/pull/582/files#diff-02a255696d374d3f496912f418e45429L61-R71

I suspected there might be similar issues with `post`, `put` and `delete` in other endpoints as well. Ran a few simple checks, and found a few mismatches. Have cross-checked with the API docs and found a few that are rightly so, but a few actual bugs as well. This PR fixes all those mismatches.


akashdotsrivastava:
@NARKOZ Need a review to merge this.

NARKOZ:
Sorry, I'll get to this today.

NARKOZ (approved):

PR #588, opened 2020-11-03
Add method to list all members including inherited members

AndreiMGagiu:
The purpose of this MR is to implement a method to list all members including inherited members. Currently in the rudy sdk we have available only two methods, that don't return the inherited ones.

- `team_members`, `group_members` - Gets a list of group or project members viewable by the authenticated user. Returns only direct members and not inherited members through ancestors groups.

Rather than trying to find a workaround(although I am not sure there is one) I think it is a good idea to have this endpoint as in the REST API that will return all members.

### DIFF
--- a/lib/gitlab/client/projects.rb
+++ b/lib/gitlab/client/projects.rb
@@ -102,6 +102,22 @@ class Gitlab::Client
       get("/projects/#{url_encode project}/members", query: options)
     end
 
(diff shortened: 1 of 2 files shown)


NARKOZ (approved):

NARKOZ:
Thank you ❤️

PR #607, opened 2021-02-24
Add changelog endpoint

andrcuns:
Adds changelog entry generation endpoint introduced in gitlab 13.9: https://docs.gitlab.com/ee/api/repositories.html#generate-changelog-data


spec/gitlab/objectified_hash_spec.rb
```diff
@@ -13,8 +13,10 @@
     let(:oh) { described_class.new(hash) }
 
     it 'allows to call Hash methods' do
-      expect(oh['foo']).to eq('bar')
-      expect(oh.merge(key: :value)).to eq('foo' => 'bar', key: :value)
+      capture_output do
+        expect(oh['foo']).to eq('bar')
+        expect(oh.merge(key: :value)).to eq('foo' => 'bar', key: :value)
+      end
```
NARKOZ, line 19: why is this change necessary?

andrcuns: I can remove this. Those expect statements are printing some stuff to stdout which is inserted in the middle of rspec test output slightly messing up the visuals, but definitely not strictly necessary if you feel this is unnecessary.

At least I think I added it because of that, it's been a while.

andrcuns: Yes, as the commit message says `Silence warning in test`

NARKOZ: Let's revert it

andrcuns:
Should be good now, removed all unrelated changes.

NARKOZ (approved):

NARKOZ:
Thanks!

PR #613, opened 2021-04-01
Allow newer versions of terminal-table

theflow:
[terminal-table](https://rubygems.org/gems/terminal-table) released a 2.0 and 3.0 which other gems starting to depend on. How do you feel about relaxing the requirement on it again here?


NARKOZ (approved):

NARKOZ:
Thank you ❤️

PR #615, opened 2021-05-07
Ruby 3 support

atog:
I ran into a problem (wrong number of arguments) passing optional params to the variables endpoint using Ruby 3.

The problem related to this change: Keyword arguments are separated from other arguments (https://www.ruby-lang.org/en/news/2020/12/25/ruby-3-0-0-released/).

### DIFF
--- a/lib/gitlab.rb
+++ b/lib/gitlab.rb
@@ -42,13 +42,6 @@ module Gitlab
   end
 
   # Delegate to Gitlab::Client
-  def self.method_missing(method, *args, &block)
-    return super unless client.respond_to?(method)
-
-    client.send(method, *args, &block)
-  end
-
-  # Delegate to Gitlab::Client
   def self.respond_to_missing?(method_name, include_private = false)
     client.respond_to?(method_name) || super
   end
--- a/lib/gitlab.rb
+++ b/lib/gitlab.rb
@@ -24,11 +24,13 @@ module Gitlab
   if RUBY_VERSION >= '3.0.0'
     def self.method_missing(method, *args, **keywargs, &block)
       return super unless client.respond_to?(method)
+
       client.send(method, *args, **keywargs, &block)
     end
   else
     def self.method_missing(method, *args, &block)
       return super unless client.respond_to?(method)
+  
       client.send(method, *args, &block)
     end
   end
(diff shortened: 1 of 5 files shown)


lib/gitlab.rb
```diff
@@ -21,11 +21,16 @@ def self.client(options = {})
...
   end
 
-  # Delegate to Gitlab::Client
-  def self.method_missing(method, *args, &block)
-    return super unless client.respond_to?(method)
-
-    client.send(method, *args, &block)
+  if RUBY_VERSION >= '3.0.0'
+    def self.method_missing(method, *args, **keywargs, &block)
+      return super unless client.respond_to?(method)
```
NARKOZ, line 26: Could you add an empty line after just to fix the rubocop offense?

lib/gitlab.rb
```diff
@@ -21,11 +21,16 @@ def self.client(options = {})
...
+      client.send(method, *args, **keywargs, &block)
+    end
+  else
+    def self.method_missing(method, *args, &block)
+      return super unless client.respond_to?(method)
```
NARKOZ, line 31: ditto

NARKOZ (changes requested):
Thank you for the PR! I'll merge it after the comments get addressed.

lib/gitlab.rb
atog: 👍🏻

lib/gitlab.rb
atog: 👍🏻

lib/gitlab.rb
```diff
@@ -21,11 +21,18 @@ def self.client(options = {})
     Gitlab::Client.new(options)
   end
 
-  # Delegate to Gitlab::Client
-  def self.method_missing(method, *args, &block)
-    return super unless client.respond_to?(method)
-
-    client.send(method, *args, &block)
+  if RUBY_VERSION >= '3.0.0'
```
asedge, line 24: I think you should use `Gem::Version` to compare the version strings. See doc [here](https://docs.ruby-lang.org/en/2.6.0/Gem/Version.html).

atog: You are right, that approach is a much better. Thanks.

NARKOZ (approved):

NARKOZ:
Thank you ❤️

PR #628, opened 2021-10-08
Respect HTTP 429 and Retry-After header

craigmiskell-gitlab:
When a 429 is received, sleep for the length of time indicated by Retry-After, then try again.

Try up to 3 times before giving up, although the count can be configured with :ratelimit_retries if you want more or less

If no Retry-After is sent by the server, defaults to 2 seconds (something is better than nothing). That should only happen with quite old and unsupported versions of GitLab

### DIFF
--- a/lib/gitlab/request.rb
+++ b/lib/gitlab/request.rb
@@ -48,7 +48,18 @@ module Gitlab
           params[:headers].merge!(authorization_header)
         end
 
-        validate self.class.send(method, endpoint + path, params)
+        retries_left = params[:ratelimit_retries] || 3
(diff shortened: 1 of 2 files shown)


craigmiskell-gitlab:
> I plan to release this week.

@NARKOZ It's been a while; any chance we could get that release soon?  I would be very grateful.

NARKOZ:
@craigmiskell-gitlab I've just released a new version and pushed to Rubygems https://github.com/NARKOZ/gitlab/releases/tag/v4.18.0

Sorry for the wait!

PR #640, opened 2022-04-13
Raise error for any 4xx or 5xx HTTP status code

### LINKED ISSUE #639: Raise error for all 4xx and 5xx response status codes
Currently, errors are raised for specific HTTP status codes ([error.rb#L142](https://github.com/NARKOZ/gitlab/blob/fcab58ca428b31d5626d9bbb24a28c63e5e8f2da/lib/gitlab/error.rb#L142-L156)). This allows us to have specially named classes like `BadRequest` (for 400 status code).

But it means that for any status code not in the map, no error is raised. This can cause subtle bugs when a 4xx or 5xx status code is received but it doesn't raise an error. This is especially problematic for APIs that don't return a JSON response, such as the [`file_contents`](https://www.rubydoc.info/gems/gitlab/Gitlab/Client/RepositoryFiles#file_contents-instance_method) method.

What do you think of raising an error for any 4xx and 5xx status code?

- If the status code has a special class associated with it, we can raise an error of that class (as happens now).
- But if there is no special class available, raise a `ResponseError`.

reuben453:
Raise error for any 4xx or 5xx HTTP status code. If there is no special error class, raise `ResponseError`.

Fixes #639

### DIFF
--- a/lib/gitlab/error.rb
+++ b/lib/gitlab/error.rb
@@ -154,5 +154,17 @@ module Gitlab
       503 => ServiceUnavailable,
       522 => ConnectionTimedOut
     }.freeze
+
+    # Returns error class that should be raised for this response. Returns nil
+    # if the response status code is not 4xx or 5xx.
+    #
+    # @param  [HTTParty::Response] response The response object.
+    # @return [Class<Error::ResponseError>, nil]
+    def self.klass(response)
+      error_klass = STATUS_MAPPINGS[response.code]
+      return error_klass if error_klass
+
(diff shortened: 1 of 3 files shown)


NARKOZ (approved):

reuben453:
Thanks @NARKOZ! Could you create a new tag please?

NARKOZ:
[Version 4.19.0](https://github.com/NARKOZ/gitlab/releases/tag/v4.19.0) was pushed to Rubygems.

PR #641, opened 2022-04-25
Add all group members endpoint

garaujodev:
As described in the [members endpoint](https://docs.gitlab.com/ee/api/members.html#list-all-members-of-a-group-or-project): 

> Returns only direct members and not inherited members through ancestors groups.

This PR introduces [all members endpoint](https://docs.gitlab.com/ee/api/members.html#list-all-members-of-a-group-or-project-including-inherited-and-invited-members):

> [...] including inherited members, invited users, and permissions through ancestor groups.

### DIFF
--- a/lib/gitlab/client/groups.rb
+++ b/lib/gitlab/client/groups.rb
@@ -71,6 +71,21 @@ class Gitlab::Client
       get("/groups/#{url_encode id}/members", query: options)
     end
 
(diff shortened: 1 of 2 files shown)


wevtimoteo (commented):
Nice job! 👏🏽

NARKOZ (approved):

NARKOZ:
Thank you ❤️

PR #646, opened 2022-08-02
Add users disable 2FA

antons-:
Adds support for the [disable two factor authentication](https://docs.gitlab.com/ee/api/users.html#disable-two-factor-authentication) endpoint introduced in GitLab `15.2`.

### DIFF
--- a/lib/gitlab.rb
+++ b/lib/gitlab.rb
@@ -50,7 +50,7 @@ module Gitlab
   # @return [Array<Symbol>]
   def self.actions
     hidden =
-      /endpoint|private_token|auth_token|user_agent|sudo|get|post|put|\Adelete\z|validate\z|request_defaults|httparty/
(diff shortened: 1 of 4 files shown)


antons-:
Regarding the failed Rubocop test - I wasn't sure on the best way to deal with this. Did we want to increase the line length limit?

asedge:
You _could_ disable the cop, via comment, for that one line.  Not sure how @NARKOZ would feel about updating the line length for the whole project.

antons-:
Thanks @asedge for the suggestion. For now I have disabled the cop via comment, and now all the tests are passing.

I should point out that the max limit for the `Layout/LineLength` cop is 127 characters - whereas my change makes the line 128 characters (1 character over the limit).

@NARKOZ Let me know what you think.

asedge (approved):
I'll approve and leave it for a few days to see if @NARKOZ has any input. After that I'll merge it in and we'll deal with it later (if needed).  Thanks so much for the contribution!

asedge:
Thank you for your contribution! 😄

PR #665, opened 2023-04-10
Implement group hooks crud

dipolesource:
This implements the CRUD API for Group Hooks https://docs.gitlab.com/ee/api/groups.html#hooks

While implementing we noticed that the CLI spec exits early, thus when rake spec is run with a random seed you do not run the full test suite. We tried setting a CI env but that didn't change the behavior. Rather than jam up this PR with extra behavior, we implemented here without updating the test suite. Happy to chat with you about submitting another PR that updates contributor doc, test harness behavior, or tests to help the issue described here.


floh96:
@dipolesource could you reopen this pr?
@NARKOZ could you reopen and review this pr please?

NARKOZ (approved):

Issue #668, opened 2023-05-30
Support using CI_JOB_TOKEN for authentication


balasankarc:
When using the job token, the header needs to be `JOB-TOKEN` instead of `PRIVATE-TOKEN`.

Starting with GitLab 14.5, `glpat-` is used as a default prefix for personal access tokens (though this is configurable). So maybe we can accept `pat_prefix` as an argument, and then detect if the specified token starts with it or not, and then set the headers accordingly?

balasankarc:
Maybe something like the following?

```diff
diff --git a/lib/gitlab/configuration.rb b/lib/gitlab/configuration.rb
index 8c37354..fa5fcc7 100644
--- a/lib/gitlab/configuration.rb
+++ b/lib/gitlab/configuration.rb
@@ -5,7 +5,7 @@ module Gitlab
   # Defines constants and methods related to configuration.
   module Configuration
     # An array of valid keys in the options hash when configuring a Gitlab::API.
-    VALID_OPTIONS_KEYS = %i[endpoint private_token user_agent sudo httparty].freeze
+    VALID_OPTIONS_KEYS = %i[endpoint private_token user_agent sudo httparty pat_prefix].freeze
 
     # The user agent that will be sent to the API endpoint if none is set.
     DEFAULT_USER_AGENT = "Gitlab Ruby Gem #{Gitlab::VERSION}"
@@ -37,6 +37,7 @@ module Gitlab
     def reset
       self.endpoint       = ENV['GITLAB_API_ENDPOINT'] || ENV['CI_API_V4_URL']
       self.private_token  = ENV['GITLAB_API_PRIVATE_TOKEN'] || ENV['GITLAB_API_AUTH_TOKEN']
+      self.pat_prefix     = 'glpat-'
       self.httparty       = get_httparty_config(ENV['GITLAB_API_HTTPARTY_OPTIONS'])
       self.sudo           = nil
       self.user_agent     = DEFAULT_USER_AGENT
diff --git a/lib/gitlab/request.rb b/lib/gitlab/request.rb
index 218b81b..307f5ee 100644
--- a/lib/gitlab/request.rb
+++ b/lib/gitlab/request.rb
@@ -12,7 +12,7 @@ module Gitlab
     headers 'Accept' => 'application/json', 'Content-Type' => 'application/x-www-form-urlencoded'
     parser(proc { |body, _| parse(body) })
 
-    attr_accessor :private_token, :endpoint
[...]

PR #669, opened 2023-05-30
Support using CI_JOB_TOKEN for authentication

balasankarc:
Closes: https://github.com/NARKOZ/gitlab/issues/668

Support specifying a prefix for personal access tokens and support using CI_JOB_TOKEN for authentication if the the token doesn't start with that prefix.

Ideally, we should default the prefix to `glpat-` so that it works out of the box. But that may break users on older GitLab instances where the prefix isn't present.

### DIFF
--- a/lib/gitlab/configuration.rb
+++ b/lib/gitlab/configuration.rb
@@ -5,7 +5,7 @@ module Gitlab
   # Defines constants and methods related to configuration.
   module Configuration
     # An array of valid keys in the options hash when configuring a Gitlab::API.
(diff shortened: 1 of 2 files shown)


balasankarc:
@NARKOZ How do you feel about something like this? Ideally, we should default the prefix to `glpat-` so that it works out of the box. But that may break users on older GitLab instances where the prefix isn't present. So, I think we can do that in a major version as a breaking change.

floh96:
@balasankarc  could you reopen this pr?
@NARKOZ could you reopen and review this pr please?

balasankarc:
New PR at https://github.com/NARKOZ/gitlab/pull/700

PR #699, opened 2024-09-10
Add get_pipelines_by_pipeline_schedule method

caiconkhicon:
# What does this PR do?

GitLab pipeline now support getting all pipelines triggered by a pipeline schedule in a project
(https://docs.gitlab.com/ee/api/pipeline_schedules.html#get-all-pipelines-triggered-by-a-pipeline-schedule).

```ruby
Gitlab.get_pipelines_by_pipeline_schedule(5, 3)
```

### DIFF
--- a/lib/gitlab/client/pipeline_schedules.rb
+++ b/lib/gitlab/client/pipeline_schedules.rb
@@ -30,6 +30,18 @@ class Gitlab::Client
       get("/projects/#{url_encode project}/pipeline_schedules/#{id}")
     end
 
+    # Get all pipelines triggered by a pipeline schedule
+    #
+    # @example
+    #   Gitlab.get_pipelines_by_pipeline_schedule(5, 3)
+    #
+    # @param  [Integer, String] project The ID or name of a project.
+    # @param  [Integer] id The ID of the pipeline schedule.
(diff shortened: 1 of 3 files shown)


lib/gitlab/client/pipeline_schedules.rb
```diff
@@ -30,6 +30,18 @@ def pipeline_schedule(project, id)
...
 
+    # Get all pipelines triggered by a pipeline schedule
+    #
+    # @example
+    #   Gitlab.get_pipelines_by_pipeline_schedule(5, 3)
+    #
+    # @param  [Integer, String] project The ID or name of a project.
+    # @param  [Integer] id The ID of the pipeline schedule.
+    # @return  [Array<Gitlab::ObjectifiedHash>]
+    def get_pipelines_by_pipeline_schedule(project, id)
```
NARKOZ, line 41: Could we rename the method so that it doesn't contain a verb?

```suggestion
    def pipelines_by_pipeline_schedule(project, id)
```

caiconkhicon: good idea. I applied you suggestion. Thanks.

PR #712, opened 2025-03-21
Add body_as_json configuration parameter

seanmil:
Allow the user to configure the client to use a Content-Type of application/json for all requests with a body unless the request explicitly provides its own Content-Type header.

Using application/json provides a broader range of compatibility with the GitLab API, especially for calls where the body expects more complex data structures.

This defaults to false for backwards compatibility.

### DIFF
--- a/lib/gitlab/configuration.rb
+++ b/lib/gitlab/configuration.rb
@@ -5,7 +5,7 @@ module Gitlab
   # Defines constants and methods related to configuration.
   module Configuration
     # An array of valid keys in the options hash when configuring a Gitlab::API.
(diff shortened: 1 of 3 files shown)


seanmil:
Note that I believe always using application/json by default would be a good change for the next non-backwards-compatible release.  I don't believe any part of the GitLab API cannot accept JSON bodies (excepting file uploads requiring multipart form-data, such as https://docs.gitlab.com/api/secure_files/#create-secure-file)

NARKOZ (approved):

NARKOZ:
Thank you!

PR #718, opened 2025-07-01
Add the set override flag API endpoint (of LDAP-sync'ed users)

katrinleinweber:
Forwarding a suggestion by a GitLab customer to add this `post("/groups/#{url_encode team_id}/members/#{user_id}/override")` method, as it proved useful for custom scripting against a self-hosted GitLab server. I've added tests and a fixture, _greening_ `rspec` locally for me.

The [API endpoind is documented here](https://gitlab.com/gitlab-org/gitlab/-/blob/v18.1.1-ee/doc/api/members.md#set-override-flag-for-a-member-of-a-group).

### DIFF
--- a/lib/gitlab/client/groups.rb
+++ b/lib/gitlab/client/groups.rb
@@ -161,6 +161,17 @@ class Gitlab::Client
       post("/groups/#{url_encode team_id}/members", body: body)
     end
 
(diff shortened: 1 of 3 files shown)


NARKOZ:
Thank you ❤️

PR #737, opened 2026-04-23
Add latest_pipeline endpoint (#656)

### LINKED ISSUE #656: Latest pipeline API method
Hi! Right now we have Gitlab.pipelines which is great, but I don't see a method which provides the "latest pipeline" API call. Reference this part of the docs:
https://docs.gitlab.com/ee/api/pipelines.html#get-the-latest-pipeline
The format is just `GET /projects/:id/pipelines/latest`

mvanhorn:
Closes #656.

Wraps the GitLab REST endpoint \`GET /projects/:id/pipelines/latest\` ([docs](https://docs.gitlab.com/ee/api/pipelines.html#get-the-latest-pipeline)).

## Usage

\`\`\`ruby
Gitlab.latest_pipeline(5)                  # latest pipeline on the project's default branch
Gitlab.latest_pipeline(5, ref: 'develop')  # latest pipeline on a specific ref
\`\`\`

The method follows the same shape as the existing \`pipeline\` / \`pipelines\` methods in \`lib/gitlab/client/pipelines.rb\` and forwards \`:ref\` (and any other future query params) through \`get\`'s \`query:\` argument so callers don't have to remember which params the endpoint supports today.

## Tests

\`spec/gitlab/client/pipelines_spec.rb\` adds two contexts:

- \"without a ref option\" — verifies the request URL and that the response is parsed into a \`Gitlab::ObjectifiedHash\`.
- \"with a ref option\" — verifies the \`ref\` query param is forwarded.

Backed by a new \`spec/fixtures/latest_pipeline.json\` fixture mirroring the existing \`pipeline.json\` shape.

I could not run \`bundle exec rspec\` locally — system Ruby is 2.6 and the gemspec requires Ruby 3.2. \`ruby -c\` confirms the source and spec parse cleanly. Happy to fix anything CI flags.


mvanhorn:
The Rubocop failure here isn't from this PR — I reproduced it on `master` at the same commit range and got the identical 6 offenses:

- `lib/gitlab/shell_history.rb:33` — `Style/FileOpen`
- `spec/gitlab/client/projects_spec.rb:549,566` and `spec/gitlab/error_spec.rb:40,46,62` — `RSpec/MatchWithSimpleRegex`

This PR only touches `lib/gitlab/client/pipelines.rb`, its spec, and a fixture, none of which are flagged. 5 of 6 are autocorrectable with `rubocop -a`. Happy to send a separate lint-only PR for them if that's useful — kept out of this one so the endpoint change stays reviewable on its own.

mvanhorn:
The Rubocop job here isn't from this PR. I ran it locally on this branch and on `master` and got the same six offences either way:

- `lib/gitlab/shell_history.rb:33` -- `Style/FileOpen`
- `spec/gitlab/client/projects_spec.rb:549,566` and `spec/gitlab/error_spec.rb:40,46,62` -- `RSpec/MatchWithSimpleRegex`

None of them are in the three files this PR touches (`lib/gitlab/client/pipelines.rb`, `spec/gitlab/client/pipelines_spec.rb`, `spec/fixtures/latest_pipeline.json`) -- it looks like a newer Rubocop picked up existing code. All four Ruby test jobs pass.

I've left them alone so this PR stays scoped to the endpoint. Happy to send a separate lint-only PR if that's useful -- five of the six are autocorrectable and the `File.open` one wants a real look.

